PR #78, opened 2021-07-14
Core refactor

shafeeqd959:


shafeeqd959:
1. remove unnecessary logic from the core
2. update command class typescript and the functionalities
3. implement configuration handler
4. contentstack utilities update
5. update plugins to support new core restructure

PR #123, opened 2021-09-21
fix: added a fix for assets not being published

abhinav-from-contentstack:
Description: 
This is a fix for the issue where assets were not getting published while using the cross publish command.
This command uses the sync api for publishing both assets and entries, the problem was that the `type` paramter, in the 
`filter` query parameter needed to be left empty for publishing both entries and assets.

Here is the relevant [ticket](https://contentstack.atlassian.net/browse/CS-18803)


shafeeqd959 (approved):

abhinav-from-contentstack:
Here is the relevant [ticket](https://contentstack.atlassian.net/browse/CS-18803)

PR #146, opened 2021-11-19
fix: reference_to must be an array

abhinav-from-contentstack:
Added a fix for an issue faced while importing entries, where `reference_to` property of a reference field should ban array
ticket id: CS-21058


shafeeqd959 (approved):

aravindbuilt (approved):

nandeesh-gajula (approved):
Approved

PR #198, opened 2022-03-08
Fixed issue - CS-23842 - Work flow import is failing

netrajpatel:
Resolved issue of CS-23842 - Workflow import is failing
 - Added info log to make the user aware that we are skipping importing workflow admin users
 - Removed workflow admin users from workflow import payload

### DIFF
--- a/packages/contentstack-import/package-lock.json
+++ b/packages/contentstack-import/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "@contentstack/cli-cm-import",
-	"version": "0.1.1-beta.11",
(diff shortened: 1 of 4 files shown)


kego1992 (commented):
are we accounting for stage transition access
we need to check and fix if it has also the issues.
because there also we add users , so this might also throw issues

kego1992:
As discussed with @netrajpatel , this issue wont arise since its taken care off.

kego1992 (approved):

shafeeqd959 (approved):

PR #201, opened 2022-03-16
fix: added a check for region data

abhinav-from-contentstack:


shafeeqd959 (commented):
bump the core version

abhinav-from-contentstack:
updated the core version

dhavaljain999 (approved):

kego1992 (approved):

PR #221, opened 2022-04-18
SRE issues for clone package

sachinkalecs:
**Unresolved issues which might impact the existing code and/or are minor severity:**
- Expected a for-of loop instead of a for loop with this simple iteration

**Horusec:**
- All are false positives

**Gitty leaks/Git Secrets:**
- All are false positives


netrajpatel (approved):
@sachinkalecs Please resolve conflicts and merge this PR

PR #234, opened 2022-04-20
clone command failing in case of multiple branches

sachinkalecs:


shafeeqd959:
@netrajpatel before merge this code we need to test this fix on this branch itself

PR #253, opened 2022-05-07
Fix: entries:migrate-html-rte security issues

hiteshshetty-dev:


shafeeqd959:
@hiteshshetty-dev , Please remove changes other than migrate-rte plugin

PR #268, opened 2022-05-12
feat: added support for pagination in export csv

vkalta:


abhinav-from-contentstack (approved):
This looks good @vkalta . I noticed the list of content types is looping even though `loop` is set to false. Could be an issue with the `inquirer-checkbox-plus-prompt` npm package.

packages/contentstack-export-to-csv/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@contentstack/cli-cm-export-to-csv",
   "description": "Export entities to csv",
-  "version": "1.0.0",
```
shafeeqd959, line 4: keep the version to 1.0.0

PR #284, opened 2022-05-25
chore: added a new command to replace publish-unpublished entries

vkalta:
Added new command to replace cm:bulk-publish:publish-unpublished


packages/contentstack-bulk-publish/src/commands/cm/bulk-publish/unpublished-entries.js
```diff
@@ -1,100 +1,15 @@
 const { Command, flags } = require('@contentstack/cli-command');
```
netrajpatel, line 1: You can remove this file and add deprecation in package.json and use alias for old command in new command

netrajpatel (changes requested):

packages/contentstack-bulk-publish/src/commands/cm/bulk-publish/unpublished-entries.js
vkalta: These changes have been made!

netrajpatel (changes requested):
@vkalta 
1. I can still see the old deprecated flags when I use --help command
2. Also the new flags does not work with command
3. The examples and help text should also be changed as per the new flags

netrajpatel (approved):

PR #291, opened 2022-06-01
content-type flag update as content-types where multiple true

sachinkalecs:


packages/contentstack-export/src/commands/cm/stacks/export.js
```diff
@@ -21,7 +21,7 @@ class ExportCommand extends Command {
     const securedAssets = exportCommandFlags['secured-assets'];
     const data = exportCommandFlags.data || exportCommandFlags['data-dir'];
     const moduleName = exportCommandFlags.module;
-    const contentTypes = exportCommandFlags['content-type'];
+    const contentTypes = exportCommandFlags['content-types'];
```
netrajpatel, line 24: This is invalid as we are passing single CT in export.

sachinkalecs: Content types in export are multiple so keeping this change

netrajpatel (approved):

PR #297, opened 2022-06-03
Fix/cs 26885

abhinav-from-contentstack:
Fix for https://contentstack.atlassian.net/browse/CS-26885
and added a fix for maintaining json rte children sequence


netrajpatel (changes requested):

packages/contentstack-import/src/lib/import/entries.js
```diff
@@ -1014,17 +1014,56 @@ importEntries.prototype = {
...
+
+                // i am facing a Maximum call stack exceeded issue, 
+                // probably because of this loop operation
+
+                let entryRefs = sourceStackEntry[ctSchema[i].uid][index].children
+                .map((e, index) => {
+                  return { index: index, value: e }
+                })
+                .filter(e => this.doEntryReferencesExist(e.value))
+                .map(e => {
```
shafeeqd959, line 1026: can't we reduce this array chanining ?

PR #303, opened 2022-06-06
Lodash dependency removed

antonyagustine:
- Removed lodash dependency


antonyagustine:
> @antonyagustine You also need to update the package versions in all package json files with updated clone plugin version where ever it is used.

@netrajpatel updated the package version as discussed.

netrajpatel (approved):

shafeeqd959 (approved):

PR #304, opened 2022-06-06
Config handler fixes

antonyagustine:
- Lodash dependency dependency added in package manager

### DIFF
--- a/packages/contentstack-utilities/src/logger.ts
+++ b/packages/contentstack-utilities/src/logger.ts
@@ -1,5 +1,5 @@
-import * as winston from 'winston';
-import config from './config-handler';
+import winston from 'winston';
+// import config from './config-handler';
 import messageHandler from './message-handler';
 
 // config.set('logger.level', 'debug');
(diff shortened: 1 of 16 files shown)


netrajpatel (commented):
@antonyagustine Though the complexity wise both ways looks simillar, using lodash.isEmpty() makes our code more readable as well as handles more cases of array as well as object with boolean return type rather than integer.
Also, we should use uniform code structure everywhere to reduce the further confusions, So I think, you should use the lodash isEmpty which is better maintainable in future as well.
I'm open for any further discussions, let me know.

antonyagustine:
> @antonyagustine Though the complexity wise both ways looks simillar, using lodash.isEmpty() makes our code more readable as well as handles more cases of array as well as object with boolean return type rather than integer. Also, we should use uniform code structure everywhere to reduce the further confusions, So I think, you should use the lodash isEmpty which is better maintainable in future as well. I'm open for any further discussions, let me know.

@netrajpatel I've reverted the code back to lodash. @shafeeqd959 As we discussed I've added lodash dependency in package manager.

packages/contentstack-utilities/src/logger.ts
```diff
@@ -1,5 +1,5 @@
-import * as winston from 'winston';
-import config from './config-handler';
+import winston from 'winston';
+// import config from './config-handler';
```
netrajpatel, line 2: Remove unused/commented code

netrajpatel (changes requested):

PR #306, opened 2022-06-07
fix(export-to-csv): fixed pagination issue with entires

vkalta:


netrajpatel (changes requested):
@vkalta Please update the package versions as well.

netrajpatel (changes requested):
@vkalta The package versions are still not updated in all dependant package and package-lock json files - 
below are the files which are still showing old versions - 
packages/contentstack/package.json
packages/contentstack-export-to-csv/package-lock.json

vkalta:
@netrajpatel Updated the versions.

netrajpatel (approved):

PR #349, opened 2022-06-23
chore: added license and security files for each plugin

vkalta:
This PR relates to CS-27613.


netrajpatel (approved):

packages/contentstack-auth/SECURITY.md
```diff
@@ -0,0 +1,27 @@
+## Security
+
+Contentstack takes the security of our software products and services seriously, which includes all source code repositories managed through our GitHub organizations.
+
```
shafeeqd959, line 4: Security files needs to be removed from the plugins, just keep it in the root

PR #416, opened 2022-07-25
feat(custom-roles): added support to import/export custom-roles

vkalta:
This relates to CS-27152


packages/contentstack-import/src/lib/import/custom-roles.js
```diff
@@ -0,0 +1,80 @@
...
+
+    if (!self.customRoles) {
+      addlogs(config, chalk.white('No custom-roles found'), 'error');
+      return;
+    }
+    self.customRolesUids = Object.keys(self.customRoles);
+    for (const uid of self.customRolesUids) {
+      const customRole = self.customRoles[uid];
+
+      if (uid in self.customRolesUidMapper) {
```
shafeeqd959, line 52: do we to loop in mapper to check it already created or not, or  self.customRolesUidMapper[uid]  o(1) operation

packages/contentstack-export/src/config/default.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
     types: [
       'stack',
       'assets',
+      'custom-roles',
```
shafeeqd959, line 12: keep the custom roles just above the workflows, is there any other module dependent on custom roles ?

packages/contentstack-import/src/config/default.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
   modules: {
     types: [
       'locales',
+      'custom-roles',
```
shafeeqd959, line 14: Keep it just above the workflows , is there any other module other than workflow dependent on the custom roles ? is there any reason that you kept custom roles after locales

shafeeqd959:
@vkalta , please resolve the comments with solution and reply for the questions. cc : @netrajpatel

packages/contentstack-import/src/lib/import/custom-roles.js
vkalta: @shafeeqd959 There is no loop in mapper object. Looping is done for keys in customRoles object.

vkalta:
@shafeeqd959 I have made the changes for moving custom-roles module just before workflows.
cc: @netrajpatel

netrajpatel (approved):

PR #421, opened 2022-07-27
fix(stack-clone): added custom abort controller api implementation to support older node versions

vkalta:
This PR relates to CS-24471.

### DIFF
--- a/packages/contentstack-clone/src/lib/util/abort-controller.js
+++ b/packages/contentstack-clone/src/lib/util/abort-controller.js
@@ -25,8 +25,6 @@ class CustomAbortSignal {
     const handlerName = `on${type}`;
 
     if (typeof this[handlerName] === 'function') this[handlerName](event);
-
-    // this.eventEmitter.emit(type, event);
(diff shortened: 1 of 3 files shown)


packages/contentstack-clone/src/lib/util/abort-controller.js
```diff
@@ -0,0 +1,51 @@
...
+  addEventListener(name, handler) {
+    this.eventEmitter.on(name, handler);
+  }
+  dispatchEvent(type) {
+    const event = { type, target: this };
+    const handlerName = `on${type}`;
+
+    if (typeof this[handlerName] === 'function') this[handlerName](event);
+
+    // this.eventEmitter.emit(type, event);
```
netrajpatel, line 29: @vkalta Avoid pushing unused/commented code

vkalta: @netrajpatel This is resolved.

netrajpatel (approved):

PR #422, opened 2022-07-28
Fix/cs 28478

abhinav-from-contentstack:
CS 28478

### DIFF
--- a/packages/contentstack-import/src/commands/cm/stacks/import.js
+++ b/packages/contentstack-import/src/commands/cm/stacks/import.js
@@ -180,4 +180,6 @@ ImportCommand.flags = {
 
 ImportCommand.aliases = ['cm:import'];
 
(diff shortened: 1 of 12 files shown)


packages/contentstack-import/src/commands/cm/stacks/import.js
```diff
@@ -180,4 +180,6 @@ ImportCommand.flags = {
 
 ImportCommand.aliases = ['cm:import'];
 
+ImportCommand.usage = 'cm:stacks:import [-c <value>] [-k <value>] [-d <value>] [-a <value>] [--module <value>] [--backup-dir <value>] [--branch <value>] [--import-webhook-status enable|disable|default]';
```
netrajpatel, line 183: @abhinav-from-contentstack  Can you update this line with new valid options for `import-webhook-status` flag?

PR #423, opened 2022-07-28
Fix:28227 sample entries along with content type creation

sachinkalecs:


packages/contentstack-migration/examples/rename-field/01-rename-field.js
```diff
@@ -47,6 +47,15 @@ module.exports = ({ migration, stackSDKInstance }) => {
...
           }
         }
+        for (let index = 0; index < 4; index++) {
+          let entry = {
+            title: `Awesome Blog ${index}`,
+            url: `/awesome-blog-${index}`,
+          }
+          try {
+            await stackSDKInstance.contentType(contentTypeUID).entry().create({ entry })
+          } catch (error) { }
```
netrajpatel, line 57: @sachinkalecs Please handle error in catch block.

PR #431, opened 2022-08-03
Fix/27997

abhinav-from-contentstack:
CS-27997

### DIFF
--- a/packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
+++ b/packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
@@ -50,13 +50,6 @@ class ExportToCsvCommand extends Command {
   }
 
   async run() {
-    if (!this.authToken) {
-      this.error(config.CLI_EXPORT_CSV_LOGIN_FAILED, {
-        exit: 2,
-        suggestions: ['https://www.contentstack.com/docs/developers/cli/authentication/'],
-      });
-    }
-
     try {
       let action;
       const {
@@ -82,6 +75,7 @@ class ExportToCsvCommand extends Command {
         case 'entries': {
           try {
             let stack;
+            let stackClient;
             let language;
             let contentTypes = [];
             const listOfTokens = configHandler.get('tokens');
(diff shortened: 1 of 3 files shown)


netrajpatel (approved):

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
```diff
@@ -192,26 +207,29 @@ class ExportToCsvCommand extends Command {
             util.write(this, listOfUsers, fileName, 'organization details');
           } catch (error) {
             if (error.message) {
-              this.log(`Error: ${error.message}`);
+              this.log(`${util.formatError(error)}`);
```
antonyagustine, line 210: @abhinav-from-contentstack this string concatenation is not necessary. We can remove it here. this.log(util.formatError(error));

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
```diff
@@ -192,26 +207,29 @@ class ExportToCsvCommand extends Command {
...
-
-            this.error('failed export content to csv');
           }
           break;
         }
       }
     } catch (error) {
       if (error.message) {
-        this.log(`Error: ${util.formatError(error)}`);
+        this.log(`${util.formatError(error)}`);
```
antonyagustine, line 218: @abhinav-from-contentstack this string concatenation is not necessary. We can remove it here. this.log(util.formatError(error));

antonyagustine (changes requested):

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
abhinav-from-contentstack: Hi, I've removed the string concatenation.

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
abhinav-from-contentstack: Hi, I've made the requested change.

antonyagustine (approved):

PR #437, opened 2022-08-04
fix: added conditions for printing success import messages

abhinav-from-contentstack:
updates

### DIFF
--- a/packages/contentstack-import/src/app.js
+++ b/packages/contentstack-import/src/app.js
@@ -52,7 +52,7 @@ exports.initial = function (configData) {
             process.exit(1);
           });
       } else {
-        let filename = path.basename(config.data);
+        let filename = config.data ? path.basename(config.data): '';
         addlogs(config, chalk.red(filename + ' Folder does not Exist'), 'error');
       }
     }
@@ -95,7 +95,7 @@ let singleImport = async (moduleName, types, config) => {
               await util.field_rules_update(config, ctPath);
             }
           }
-          if (!data) {
+          if (!(data && data.empty)) {
             addlogs(config, moduleName + ' imported successfully!', 'success');
           }
           addlogs(config, 'The log for this is stored at ' + path.join(config.oldPath, 'logs', 'import'), 'success');
--- a/packages/contentstack-import/src/app.js
+++ b/packages/contentstack-import/src/app.js
@@ -87,7 +87,7 @@ let singleImport = async (moduleName, types, config) => {
       let exportedModule = require('./lib/import/' + moduleName);
       exportedModule
         .start(config)
-        .then(async function () {
+        .then(async function (data) {
           if (moduleName === 'content-types') {
             let ctPath = path.resolve(config.data, config.modules.content_types.dirName);
             let fieldPath = path.join(ctPath + '/field_rules_uid.json');
@@ -95,7 +95,9 @@ let singleImport = async (moduleName, types, config) => {
               await util.field_rules_update(config, ctPath);
             }
           }
-          addlogs(config, moduleName + ' imported successfully!', 'success');
+          if (!data) {
+            addlogs(config, moduleName + ' imported successfully!', 'success');
+          }
           addlogs(config, 'The log for this is stored at ' + path.join(config.oldPath, 'logs', 'import'), 'success');
           return resolve();
         })
--- a/packages/contentstack-import/src/app.js
+++ b/packages/contentstack-import/src/app.js
@@ -52,7 +52,7 @@ exports.initial = function (configData) {
             process.exit(1);
           });
       } else {
-        let filename = config.data ? path.basename(config.data): '';
+        let filename = path.basename(config.data);
         addlogs(config, chalk.red(filename + ' Folder does not Exist'), 'error');
       }
     }
--- a/packages/contentstack-import/src/lib/import/assets.js
+++ b/packages/contentstack-import/src/lib/import/assets.js
@@ -60,9 +60,9 @@ importAssets.prototype = {
     mkdirp.sync(mapperDirPath);
 
     return new Promise(function (resolve, reject) {
-      if (self.assets === undefined) {
+      if (self.assets === undefined || Object.keys(self.assets).length === 0) {
         addlogs(config, 'No Assets Found', 'success');
-        return resolve();
+        return resolve({empty: true});
       }
       let assetUids = Object.keys(self.assets);
       let batches = [];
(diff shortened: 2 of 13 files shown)


packages/contentstack-import/src/app.js
```diff
@@ -87,15 +87,17 @@ let singleImport = async (moduleName, types, config) => {
...
+        .then(async function (data) {
           if (moduleName === 'content-types') {
             let ctPath = path.resolve(config.data, config.modules.content_types.dirName);
             let fieldPath = path.join(ctPath + '/field_rules_uid.json');
             if (fieldPath && fieldPath !== undefined) {
               await util.field_rules_update(config, ctPath);
             }
           }
-          addlogs(config, moduleName + ' imported successfully!', 'success');
+          if (!data) {
```
antonyagustine, line 98: @abhinav-from-contentstack Can you check for the specific key to check the condition? instead of **if (!data)** check **if (!(data && data.empty))**.

the reason is basically it's returning a promise of unknown data all we know is it's a promise<unknown|any> data. In the feature, if someone wants to return any data for any of the functionality it'll break the above code/condition.

CC @netrajpatel

packages/contentstack-import/src/lib/import/assets.js
```diff
@@ -60,9 +60,9 @@ importAssets.prototype = {
     mkdirp.sync(mapperDirPath);
 
     return new Promise(function (resolve, reject) {
-      if (self.assets === undefined) {
+      if (self.assets === undefined || Object.keys(self.assets).length === 0) {
```
antonyagustine, line 63: @abhinav-from-contentstack Can you reverify this condition also?

it checks that **if (some_variable === undefined || Object.keys(some_variable).length === 0)** right. Ultimately if the first condition fails it's going to jump into or condition **(Object.keys(some_variable).length === 0)** to validate right, in that case it's going to fail. Because **Object.keys(undefined).length** is not a valid code.

What if my **some_variable(asset/any)** is ideally undefined, All we know is variables can be undefined as well. So better handle it to prevent such an issue.

packages/contentstack-import/src/lib/import/content-types.js
```diff
@@ -82,6 +83,10 @@ importContentTypes.prototype = {
     });
 
     return new Promise(function (resolve, reject) {
+      if (self.contentTypes === undefined || Object.keys(self.contentTypes).length === 0) {
```
antonyagustine, line 86: @abhinav-from-contentstack Can you reverify this condition once?

it checks that **if (some_variable === undefined || Object.keys(some_variable).length === 0)** right. Ultimately if the first condition fails it's going to jump into or condition **(Object.keys(some_variable).length === 0)** to validate right, in that case it's going to fail. Because **Object.keys(undefined).length** is not a valid code.

What if my **some_variable(asset/any)** is ideally undefined, All we know is variables can be undefined as well. So better handle it to prevent such an issue.

packages/contentstack-import/src/lib/import/environments.js
```diff
@@ -48,9 +48,9 @@ importEnvironments.prototype = {
 
     mkdirp.sync(envMapperPath);
     return new Promise(function (resolve, reject) {
-      if (self.environments === undefined) {
+      if (self.environments === undefined || Object.keys(self.environments).length === 0) {
```
antonyagustine, line 51: @abhinav-from-contentstack Can you reverify this condition once?

packages/contentstack-import/src/lib/import/extensions.js
```diff
@@ -52,9 +52,9 @@ importExtensions.prototype = {
     mkdirp.sync(extMapperPath);
 
     return new Promise(function (resolve, reject) {
-      if (self.extensions === undefined) {
+      if (self.extensions === undefined || Object.keys(self.extensions).length === 0) {
```
antonyagustine, line 55: @abhinav-from-contentstack Can you reverify this condition once?

packages/contentstack-import/src/lib/import/global-fields.js
```diff
@@ -62,10 +62,10 @@ importGlobalFields.prototype = {
     }
     client = stack.Client(config);
     return new Promise(function (resolve, reject) {
-      if (self.globalfields === undefined) {
+      if (self.globalfields === undefined || Object.keys(self.globalfields).length === 0) {
```
antonyagustine, line 65: @abhinav-from-contentstack Can you reverify this condition once?

packages/contentstack-import/src/lib/import/labels.js
```diff
@@ -49,9 +49,9 @@ importLabels.prototype = {
     labelFailsPath = path.resolve(config.data, 'labels', 'fails.json');
     mkdirp.sync(labelMapperPath);
     return new Promise(function (resolve, reject) {
-      if (self.labels == undefined) {
-        addlogs(config, chalk.white('No Label Found'), 'error');
-        return resolve();
+      if (self.labels == undefined || Object.keys(self.labels).length === 0) {
```
antonyagustine, line 52: @abhinav-from-contentstack Can you reverify this condition once?

packages/contentstack-import/src/lib/import/locales.js
```diff
@@ -52,9 +52,9 @@ importLanguages.prototype = {
     }
 
     return new Promise(function (resolve, reject) {
-      if (self.languages === undefined) {
-        addlogs(config, chalk.white('No Languages Found'), 'error');
-        return resolve();
+      if (self.languages === undefined || Object.keys(self.languages).length === 0) {
```
antonyagustine, line 55: @abhinav-from-contentstack Can you reverify this condition once?

packages/contentstack-import/src/lib/import/webhooks.js
```diff
@@ -49,9 +49,9 @@ importWebhooks.prototype = {
     mkdirp.sync(webMapperPath);
 
     return new Promise(function (resolve, reject) {
-      if (self.webhooks == undefined) {
+      if (self.webhooks == undefined || Object.keys(self.webhooks).length === 0) {
```
antonyagustine, line 52: @abhinav-from-contentstack Can you reverify this condition once?

packages/contentstack-import/src/lib/import/workflows.js
```diff
@@ -50,8 +50,8 @@ importWorkflows.prototype = {
     mkdirp.sync(workflowMapperPath);
     return new Promise(function (resolve, reject) {
       if (self.workflows == undefined || Object.keys(self.workflows).length === 0) {
```
antonyagustine, line 52: @abhinav-from-contentstack Can you reverify this condition once?

antonyagustine (changes requested):

antonyagustine (approved):

PR #466, opened 2022-08-29
fix(auth): skipped validate api key check to support adding mgmt tokens

vkalta:
This PR resolves CS-29665 and CS-29967.


netrajpatel:
@vkalta As this is the production bugfix, this fix should be merged in development branch.

PR #469, opened 2022-08-30
skipped validate api key check to support adding mgmt tokens

vkalta:
This PR resolves CS-29665 and CS-29967.


netrajpatel (changes requested):

shafeeqd959:
@vkalta, user should be able add management token without login into the CLI. Please logout before testing these changes. Here management client is being initiated using "this.authtoken" that will throw an error in case of logged out user. We have to initiate management API client using management token, then stack using API key. For more clarification, we can connect. Ping me on slack

CC: @netrajpatel

shafeeqd959:
@vkalta , Please connect with on this, before proceeding it further

CC: @netrajpatel

vkalta:
@shafeeqd959 @netrajpatel All the changes requested are added and custom region issue is also fixed.

PR #474, opened 2022-09-05
fix(custom-roles): added support to migrate locales,entries and env in rules

vkalta:
This PR fixes CS-29843.

### DIFF
--- a/packages/contentstack-export/src/lib/export/locales.js
+++ b/packages/contentstack-export/src/lib/export/locales.js
@@ -12,7 +12,6 @@ const helper = require('../util/helper');
 const { addlogs } = require('../util/log');
 let config = require('../../config/default');
 let localeConfig = config.modules.locales;
-const masterLocale = config.master_locale;
 let requiredKeys = localeConfig.requiredKeys;
 let stack = require('../util/contentstack-management-sdk');
 let client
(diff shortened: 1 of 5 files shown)


packages/contentstack-export/src/lib/export/locales.js
```diff
@@ -12,19 +12,14 @@ const helper = require('../util/helper');
...
 function ExportLocales() {
   this.qs = {
     include_count: true,
     asc: 'updated_at',
-    query: {
-      code: {
-        $nin: [masterLocale.code],
-      },
-    },
+    query: {},
```
netrajpatel, line 22: Why removed the condition to not fetch the master local?
We do not export the master local for any stack.
Please discuss with @shafeeqd959 for this change.

netrajpatel (changes requested):
@shafeeqd959 Can you please review this PR?

packages/contentstack-export/src/lib/export/locales.js
vkalta: @netrajpatel This condition is handled across import/export script. We can connect and walk through the code.
We need to store uid of master locale from the source stack in order to properly migrate custom roles and locales.
cc: @shafeeqd959

netrajpatel: @vkalta It will be good if we make changes of adding/deleting locals in the 'custom roles' lib file rather than changing the core logic it has lot of other dependancies.
CC: @shafeeqd959

vkalta:
Closing this PR as the changes had to be reverted.

PR #524, opened 2022-11-02
Import stability

shafeeqd959:


packages/contentstack-import/src/lib/import/entries.js
```diff
@@ -42,8 +42,10 @@ let masterLanguage;
 
 let skipFiles = ['__master.json', '__priority.json', 'schema.json'];
 let entryBatchLimit = config.rateLimit || 10;
+const importConcurrency = eConfig.importConcurrency || config.importConcurrency;
+const writeConcurrency = eConfig.writeConcurrency || config.writeConcurrency;
 
-function importEntries() {
+function EntriesImport() {
```
antonyagustine, line 48: Are we not changing entries module from functional programming to OOPs.?

antonyagustine (changes requested):

packages/contentstack-import/src/lib/import/entries.js
shafeeqd959: No really required for now

packages/contentstack-import/src/config/default.js
```diff
@@ -24,11 +24,11 @@ module.exports = {
       'marketplace-apps',
       'global-fields',
       'content-types',
-      'workflows',
+      // 'workflows',
```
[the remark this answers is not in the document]
shafeeqd959: uncommented

PR #538, opened 2022-11-09
refactor: updated versions

abhinav-from-contentstack:
Updated version for import, export, clone and cli packages


packages/contentstack/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@contentstack/cli",
   "description": "Command-line tool (CLI) to interact with Contentstack",
-  "version": "1.2.0",
+  "version": "1.2.2",
```
antonyagustine, line 4: @abhinav-from-contentstack  Why we updating this version directly **1.2.0** to **1.2.2**.? If it happend by mistake let's fix it.

CC @netrajpatel

antonyagustine (changes requested):

packages/contentstack/package.json
abhinav-from-contentstack: I updated the version to 1.2.2, because our npm package is already at 1.2.1.

shafeeqd959 (approved):

antonyagustine (approved):

PR #539, opened 2022-11-10
Ask Prompt to get encryption key

antonyagustine:
- Ask prompt to get credentials to encryption key while exporting/importing (as shared/distributed key among the users who work with same or different stack/org)
- Provide option to pass base url and encryption key through config.
- When using --yes flag must flexible to pass encrypetion and base url through config else they can pass those values when prompting them to provide


shafeeqd959:
@antonyagustine can we update the sprint 23 branch and this feature branch with the latest changes from development branch ?

CC: @netrajpatel

packages/contentstack-export/src/app.js
```diff
@@ -139,20 +138,26 @@ const allExport = async (config, types, branchName) => {
...
-      addlogs(
-        config,
-        chalk.red('Failed to migrate stack: ' + (config.sourceStackName || config.source_stack) + '. Please check error logs for more info'),
-        'error',
-      );
-      addlogs(config, chalk.red(error && error.errorMessage), 'error');
-      addlogs(config, 'The log for this is stored at ' + path.join(config.data, 'logs', 'export'), 'error');
-      return reject(error);
+      config,
+      chalk.green(
```
shafeeqd959, line 142: @antonyagustine  we have made changes in these files in latest stability release, can you make sure those are part of sprint23 and this feature branch to avoid replacing those changes

CC: @netrajpatel

antonyagustine: Yes, It had some merge conflict, That's why this code still was there. We already merged the development source with sprint23 branch.

shafeeqd959: @antonyagustine  Why did we resolve this conversation ?, if we had merge conflict then we should have kept the incoming changes. It seems incoming changes are ignored and those changes are mandatory to handle the errors in export command. Please do compare the changes in development branch with your feature branch. Make only necessary changes.

CC: @netrajpatel

antonyagustine: @shafeeqd959 I was working on some task which was assigned to me related to the marketplace. While testing I encountered this issue and added a fix on my end. And I assumed that it may be happened because of a merge conflict because most of the code seems the same. If you wanna compare with feature branches please do so. I'm currently engaged/focused on the current task scope. I do not wanna mingle/complete the problem statement. If required I can revert these changes from the app.js file. This PR focuses on marketplace app-related features which I added. If you need can have a new ticket and PR to validate this. I'm closing this conversation here.

Thanks

CC @netrajpatel

netrajpatel (approved):

PR #553, opened 2022-11-18
Fix/users pagination

antonyagustine:
- This PR fixes CS-30885 ([#546](https://github.com/contentstack/cli/pull/546))

### DIFF
--- a/packages/contentstack-export-to-csv/src/util/index.js
+++ b/packages/contentstack-export-to-csv/src/util/index.js
@@ -383,7 +383,7 @@ function getOrgUsers(managementAPIClient, orgUid, ecsv) {
   return new Promise((resolve, reject) => {
     managementAPIClient
       .getUser({ include_orgs_roles: true })
-      .then((response) => {
+      .then(async (response) => {
         let organization = response.organizations.filter((org) => org.uid === orgUid).pop();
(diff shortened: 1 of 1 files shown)


packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -396,12 +396,34 @@ function getOrgUsers(managementAPIClient, orgUid, ecsv) {
...
+async function getUsers(organization, params, result = []) {
+  try {
+    const users = await organization.getInvitations(params);
+    if (!users.items || (users.items && !users.items.length)) {
+      return result;
+    } else {
+      result = result.concat(users.items);
+      params.skip = params.page * params.limit;
+      params.page++;
+      await wait(200);
```
shafeeqd959, line 419: Is there any specific reason why added this delay ?, If so we need to consider having a configuration where we keep the delay.

antonyagustine: @shafeeqd959 the following was the @vkalta answer when it was asked earlier. `This is intended to not overload the server with requests.`

@vkalta take please it forward for any further queries.

CC @netrajpatel

vkalta (approved):

netrajpatel (approved):

PR #561, opened 2022-11-28
release workflow

shafeeqd959:


netrajpatel (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,185 @@
...
+    branches: [main]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '16.x'
```
antonyagustine, line 14: @shafeeqd959 Just wanted to confirm that are we going with node V16.x.? Because I remember we decided to support V14.x and above.

CC @netrajpatel

.github/workflows/release.yml
```diff
@@ -0,0 +1,185 @@
+name: Release CLI Plugins
+
+on:
+  push:
+    branches: [main]
```
antonyagustine, line 5: @shafeeqd959 Quick query, As per the current release plan we publish from dev and we merging to main. I can see the main branch here. In that case, we'll merge/push final changes to the main branch directly and workflow will take care of the release.?

Currently, If something goes wrong we'll test and verify and publish again fix version, then we merge it to the main branch. We trying to make sure the main branch always should have proper fix/working code I guess

CC @netrajpatel

.github/workflows/release.yml
shafeeqd959: Yes, we should support from 14.x onwards and that should be tested from the test suite itself. This is an environment that we setup for the release.

.github/workflows/release.yml
shafeeqd959: We always release the code of main and that should be proper. To make sure the stability of the code, we are updating the test suite which is listening on the dev branch changes.

.github/workflows/release.yml
antonyagustine: Why not V14.x here as well.? To make sure it works everywhere till the moment it publishes like a checkpoint.?

CC @netrajpatel

shafeeqd959: @antonyagustine , it's a valid point, this might not be the place to test these things, rather we should have tested all the major versions which we support before the code reach here. Let's see if that is not happening, then will reduce it to 14. thanks

antonyagustine (approved):

PR #584, opened 2022-12-21
Consumed utility based CMA SDK object in exports module

netrajpatel:


packages/contentstack-export/src/app.js
```diff
@@ -15,7 +15,7 @@ exports.initial = async (config) => {
     exports.getConfig = () => {
       return config;
     };
-    const APIClient = stack.Client(config);
+    const APIClient = await managementSDKClient(config);
```
antonyagustine, line 18: @netrajpatel We are not using any async functionality in the managementSDKClient; can we remove async and use it as a simple function instead of waiting for it?

netrajpatel: @antonyagustine We have to because the latest SDK version has introduced a promised call for a retry connection event in case of 401, to handle that in the future we have to make use of await or promise to get the managementSDKClient.
So by doing this, we will be making sure to reduce as many SDK connections and share the same connection instance whenever required, as well as those will be used as a promise function.

netrajpatel: Let me know your thoughts.

antonyagustine: Sure we can keep it if we have a plan for it in future.

antonyagustine (approved):

shafeeqd959 (approved):

shafeeqd959:
@netrajpatel, @antonyagustine changes seems to have formatting  changes in many file, any reasons ?. Are we using the prettier ?. Seems existing code is not formatted accordingly. Please confirm

PR #585, opened 2023-01-02
Export rewrite

shafeeqd959:


netrajpatel (approved):

packages/contentstack-export/src/interfaces/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export interface ContentStackManagementClient {
+  contentstackClient: object;
+}
+
+export interface PrintOptions {
+  color?: string;
+}
+
+export interface InquirePayload {
```
antonyagustine, line 14: @shafeeqd959 Can you recheck the InquirePayload types from their package? One of my observation is **_validate_** option is missing from the list I could see.

PR #597, opened 2023-01-20
Import rewrite: assets

antonyagustine:
**Import rewrite**
- Assets rewritten to TS and unit test cases added


packages/contentstack-import/src/commands/cm/stacks/import.ts
```diff
@@ -94,7 +94,24 @@ export default class ImportCommand extends Command {
     // initialize the importer
     // start import
     try {
-      this.managementAPIClient = { host: this.cmaHost, authtoken: this.authToken };
+      this.managementAPIClient = {
+        retryLimit: 3,
+        host: this.cmaHost,
+        authtoken: this.authToken,
+        retryCondition: (error) => {
```
netrajpatel, line 101: @antonyagustine This is already implemented in common utility-based SDK object creation.
Why we are introducing this again?

packages/contentstack-import/src/import/modules/base-class.ts
```diff
@@ -0,0 +1,218 @@
...
+          } else if (apiParams) {
+            apiParams.apiData = element;
+            promise = this.makeAPICall(apiParams, isLastRequest);
+          }
+
+          allPromise.push(promise);
+        }
+
+        /* eslint-disable no-await-in-loop */
+        await Promise.allSettled(allPromise);
```
netrajpatel, line 125: @antonyagustine Can we have a provision to handle the errors caught from this Promise call?

netrajpatel (changes requested):

packages/contentstack-import/src/commands/cm/stacks/import.ts
antonyagustine: @netrajpatel Yes, It is supposed to be used from the utility. But still, it is consumed from Command Class. And For testing purposes, I enable it.

@shafeeqd959 is working on the structure level including this change/fix. Once his code has the change I'll pull and update it here. Or I can merge the code with his PR with those changes.

packages/contentstack-import/src/import/modules/base-class.ts
antonyagustine: @netrajpatel We capture all the errors in each request. Here you can find the link for error handling.
https://github.com/contentstack/cli/pull/597/files#diff-0c1e791feb0ad5ec979aadf45f8669196546cc91d3140ec6c7fd5d70d01be9e2R180

netrajpatel (approved):

PR #615, opened 2023-02-03
Marketplace apps: OAuth redirect URL issue fixed

antonyagustine:
**Tickets: CS-32684, CS-34614**
- Code clean
- MKT apps OAuth Redirect URL issue fix
- HTTP client type definition correction, new constructor params introduced.

### DIFF
--- a/packages/contentstack-import/src/lib/import/marketplace-apps.js
+++ b/packages/contentstack-import/src/lib/import/marketplace-apps.js
@@ -387,7 +387,7 @@ module.exports = class ImportMarketplaceApps {
 
       if (installation.installation_uid) {
         log(this.config, `${app.manifest.name} app installed successfully.!`, 'success');
-        await this.makeRedirectUrlCall(installation);
+        await this.makeRedirectUrlCall(installation, app.manifest.name);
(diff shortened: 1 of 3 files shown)


packages/contentstack-import/src/lib/import/marketplace-apps.js
```diff
@@ -72,7 +74,7 @@ module.exports = class ImportMarketplaceApps {
     // NOTE get org uid
     if (this.config.auth_token) {
       const stack = await this.stackAPIClient.fetch().catch((error) => {
-        log(this.config, formatError(error), 'success');
+        log(this.config, formatError(error), 'succeiiss');
```
netrajpatel, line 77: Is this a typo?

netrajpatel (changes requested):

packages/contentstack-import/src/lib/import/marketplace-apps.js
antonyagustine: Yes, it was a typo, Fixed it.

netrajpatel (approved):

PR #619, opened 2023-02-08
Custom name for a master language doesn't get imported or cloned

aman19K:


packages/contentstack-export/src/lib/export/locales.js
```diff
@@ -10,11 +10,6 @@ class LocaleExport {
     this.qs = {
       include_count: true,
       asc: 'updated_at',
-      query: {
```
netrajpatel, line 13: @aman19K This condition was added intentionally to avoid exporting the master locale, and removing this code will affect many places.

netrajpatel (changes requested):

aman19K:
created new PR

PR #628, opened 2023-02-16
test(cli-import): added auth token test cases

vkalta:
This PR addresses CS-29676.


antonyagustine (approved):
Created a tracking ticket (CS-37077) for the enhancement and the review points implementation. So, merging this PR now

PR #636, opened 2023-03-08
Marketplace apps: private app creation fixes

netrajpatel:
Issue - CS-35811
Package versions are already updated, no need to update again in this PR.


antonyagustine (approved):

netrajpatel:
> Are we planning to use the same branch for the private app wrong name logging issue.?

@antonyagustine No, that has a new JIRA created.

PR #659, opened 2023-03-21
Feat/cs 36414

abhinav-from-contentstack:
- includes changes requested by SRE team in ticket CS-36414


abhinav-from-contentstack:
> @abhinav-from-contentstack Any version updates required?

I don't think so netraj. Because as part of the release, all packages are being updated to the next major version.

netrajpatel (approved):

PR #709, opened 2023-04-12
Merge Feat/oauth stable to development

netrajpatel:


packages/contentstack-auth/src/commands/auth/login.ts
```diff
@@ -21,42 +29,61 @@ export default class LoginCommand extends Command {
...
+      const user: User = await authHandler.login(username, password);
+      if (typeof user !== 'object' || !user.authtoken || !user.email) {
+        throw new CLIError('Failed to login - invalid response');
+      }
+      await oauthHandler.setConfigData('basicAuth', user);
+      logger.info('successfully logged in');
+      cliux.success('CLI_AUTH_LOGIN_SUCCESS');
+    } catch (error) {
+      console.error(error);
+      throw error;
```
antonyagustine, line 86: Suggestion: Do we really need to throw the error and catch it again and throw? Can we just stop the process? Because if authentication failed user can't use CLI right?

antonyagustine (approved):

PR #725, opened 2023-04-19
Feat/cs 37363

netrajpatel:
Issue - 37363


packages/contentstack-auth/src/commands/auth/logout.ts
```diff
@@ -60,9 +60,22 @@ export default class LogoutCommand extends Command {
...
+        if (error.message) {
+          if (error.message.message) {
+            errorMessage = error.message.message;
+          } else {
+            errorMessage = error.message;
+          }
+        } else {
+          errorMessage = error;
+        }
+      }
```
antonyagustine, line 74: Suggestion: This particular part of the code is repeating in both login in and logout, It can be reused [**Minor**]

https://github.com/contentstack/cli/pull/725/files#diff-a01e7a0777da625b321b5da6d541e83c12ffcb2284a5f0a556c911beb449d328R67-R78

packages/contentstack/bin/run
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-require('dotenv').config();
+require('dotenv').config({ path: '../../.env' });
```
antonyagustine, line 2: Do we need env to be part of the run file only dev is enough I guess.? Because this file will be used by the production build. So in that case we need to make `dotenv` as a main dependency, and that is not necessary if we don't need any .env file existing in prod/GA CLI. Also, prod CLI can't find the mentioned path.!

antonyagustine (approved):

packages/contentstack/bin/run
antonyagustine: @netrajpatel

netrajpatel: @antonyagustine Resolved.

packages/contentstack-utilities/src/auth-handler.ts
```diff
@@ -37,14 +39,10 @@ class AuthHandler {
 
   constructor() {
     this.codeVerifier = crypto.pseudoRandomBytes(32).toString('hex');
-    this.OAuthBaseURL = process.env.OAUTH_APP_BASE_URL || '';
-    this.OAuthAppId = process.env.OAUTH_APP_ID || '';
-    this.OAuthClientId = process.env.OAUTH_CLIENT_ID || '';
-    this.OAuthRedirectURL = process.env.OAUTH_APP_REDIRECT_URL || '';
+    this.OAuthAppId = process.env.OAUTH_APP_ID || '6400aa06db64de001a31c8a9';
+    this.OAuthClientId = process.env.OAUTH_CLIENT_ID || 'Ie0FEfTzlfAHL4xM';
+    this.OAuthRedirectURL = process.env.OAUTH_APP_REDIRECT_URL || 'http://localhost:8184';
```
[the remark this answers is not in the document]
netrajpatel: Yes @antonyagustine These hardcoded values are intentionally added.

packages/contentstack-auth/src/commands/auth/logout.ts
netrajpatel: Sure, we will change this in next commits of error handling

PR #734, opened 2023-04-24
Handle get base branch error & config changes

aman19K:


packages/contentstack-branch/src/config/index.ts
```diff
@@ -1,9 +1,8 @@
 const config = {
-  baseUrl: 'https://dev16-branches.csnonprod.com/v3/stacks/branches/compare',
+  baseUrl: 'https://dev16-api.csnonprod.com/v3/stacks/branches_compare',
```
netrajpatel, line 2: @aman19K @shafeeqd959 Why are we still using this config?
Can't we set the region and get this from the region config?

netrajpatel (changes requested):

packages/contentstack-branch/src/config/index.ts
aman19K: @netrajpatel We aren't using the configuration file in the SDK integration PR.
PR:- https://github.com/contentstack/cli/pull/732

PR #735, opened 2023-04-25
Fix/cs 37794

netrajpatel:
Multiple bugs fixed 
cs-37794
CS-37546
CS-37839
Logout issue
Readme updates

### DIFF
--- a/packages/contentstack-utilities/src/auth-handler.ts
+++ b/packages/contentstack-utilities/src/auth-handler.ts
@@ -119,14 +119,15 @@ class AuthHandler {
           if (queryObject.code) {
             cliux.print('Auth code successfully fetched.');
             this.getAccessToken(queryObject.code)
-              .then(() => {
+              .then(async () => {
+                await this.setOAuthBaseURL();
                 cliux.print('Access token fetched using auth code successfully.');
                 cliux.print(
-                  'You can review the access permissions on the page - https://app.contentstack.com/#!/marketplace/authorized-apps',
+                  `You can review the access permissions on the page - ${this.OAuthBaseURL}/#!/marketplace/authorized-apps`,
                 );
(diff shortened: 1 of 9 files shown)


packages/contentstack-utilities/src/auth-handler.ts
```diff
@@ -378,6 +384,18 @@ class AuthHandler {
...
+    return configHandler.get(this.authorisationTypeKeyName) ? configHandler.get(this.authorisationTypeKeyName) : false;
+  }
+
+  async isAuthorisationTypeBasic(): Promise<boolean> {
+    return configHandler.get(this.authorisationTypeKeyName) === this.authorisationTypeAUTHValue ? true : false;
+  }
+
+  async isAuthorisationTypeOAuth(): Promise<boolean> {
+    return configHandler.get(this.authorisationTypeKeyName) === this.authorisationTypeOAUTHValue ? true : false;
+  }
```
antonyagustine, line 397: Suggestion: All the above functions try to access from config(Means accessing file system every function call). Can we have a local state manage those values and return in the function? Unless/until changes happened in config it does need not to access config. [MINOR]

antonyagustine (approved):

packages/contentstack-utilities/src/auth-handler.ts
shafeeqd959: @antonyagustine , config library internally managing this state as per my understanding from the library. So i think we dont need this handling here.

shafeeqd959 (approved):

packages/contentstack-utilities/src/auth-handler.ts
netrajpatel: @antonyagustine The Conf package internally handles this.

PR #768, opened 2023-05-05
Global fields are not exporting in Custom regions

aman19K:


abhinav-from-contentstack:
Hi @aman19K , the version for cli-cm-export has been bumped twice (it was already bumped in #756). I'll be raising a new PR for correcting versions shortly.

PR #774, opened 2023-05-09
Bugfix-CS-37564

cs-raj:
The set custom host and region command is not working for blizzard. Due to Regex checking, fixed that


netrajpatel:
@cs-raj There are some GitHub workflows failing in this PR, can you please fix those first?

cs-raj:
Yes

netrajpatel (approved):

PR #776, opened 2023-05-09
Hotfix/launch url fix

antonyagustine:


packages/contentstack-utilities/src/auth-handler.ts
```diff
@@ -377,7 +377,7 @@ class AuthHandler {
     });
   }
 
-  async isAuthenticated(): Promise<boolean> {
+  isAuthenticated(): boolean {
```
netrajpatel, line 380: @antonyagustine Have you checked all the places where isAuthenticated used?
Whether it is with promise or non-promise function calls?

abhinav-from-contentstack (approved):

packages/contentstack-utilities/src/auth-handler.ts
antonyagustine: Yes, I verified. Only one location was a promised call. It was in the seed package, That too changes were made by me a few weeks back because I got compile error.

Now everything is corrected and verified. I will ask @AniketDev7 to run QA on dev once to make sure no issues exist.

PR #794, opened 2023-05-17
CS-37773- The user was unable to add the management token in CLI@v1.5.0

cs-raj:
https://github.com/contentstack/cli/pull/795 -- This PR solved the issue which was mentioned.


cs-raj:
The issue was resolved in https://github.com/contentstack/cli/pull/795 .

PR #819, opened 2023-05-29
Bugfix/launch fixes

akashbhosale97:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=bYs6OMX)


packages/contentstack-launch/test/unit/commands/z-functions.test.ts
```diff
@@ -0,0 +1,54 @@
...
+      assert.calledWithExactly(logStub, `Serving on port 4050`);
+    }
+    processStub.restore();
+    logStub.restore();
+    app.listen().close((err) => {
+      console.log('closing server');
+      process.exit(err ? 1 : 0);
+    });
+  });
+});
```
antonyagustine, line 54: @akashbhosale97 Is this file pushed intentionally or by mistake? I could see two files for the functions command.

packages/contentstack-launch/env-unit-tests-sample.txt
```diff
@@ -0,0 +1,3 @@
+ENVIRONMENT=<DEPLOYMENT_ENVIRONMENT>
+ORG=<ORG_UID>
+PROJECT=<PROJECT_NAME>
```
antonyagustine, line 3: Can we rename this file as `example.env`?

packages/contentstack-launch/test/unit/commands/z-functions.test.ts
akashbhosale97: it is intentional,  there is only one file just the file name was changed

packages/contentstack-launch/env-unit-tests-sample.txt
akashbhosale97: renamed

packages/contentstack-launch/test/unit/commands/z-functions.test.ts
antonyagustine: Any specific reason to have the filename prefixed with `z-` .?

akashbhosale97: the functions command has express server which keeps on running and wont close, if close then and there then next test commands wont run so changed it so that all test cases run and this one to run at last and then server can be closed

antonyagustine (approved):

netrajpatel (approved):

PR #832, opened 2023-06-09
fix(clone): fixed undo command for import flow

vkalta:
This PR fixes CS-39141.

### DIFF
--- a/packages/contentstack-clone/src/lib/util/clone-handler.js
+++ b/packages/contentstack-clone/src/lib/util/clone-handler.js
@@ -1,12 +1,12 @@
 const ora = require('ora');
 const path = require('path');
 const inquirer = require('inquirer');
-const rimraf = require('rimraf');
 const chalk = require('chalk');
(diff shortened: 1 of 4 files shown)


packages/contentstack-clone/src/lib/util/clone-handler.js
```diff
@@ -1,12 +1,12 @@
 const ora = require('ora');
 const path = require('path');
 const inquirer = require('inquirer');
-const rimraf = require('rimraf');
 const chalk = require('chalk');
+const prompt = require('prompt');
+const colors = require('@colors/colors/safe');
```
netrajpatel, line 6: @vkalta I saw, you have used colors package items in the code but I don't see the colors package in package.json, can you add that otherwise this can be failed while installing and releasing the individual package.

netrajpatel (changes requested):

netrajpatel (approved):

PR #844, opened 2023-06-19
Fix export to CSV flags and unit tests

akashbhosale97:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=lLUjxkr)

FIxed flags : 
1.  --org-name
2.  --stack-name
3.  --content-type

New flags :
1.  --stack-api-key

Unit Tests added for export to csv command


packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
```diff
@@ -33,6 +33,12 @@ class ExportToCsvCommand extends Command {
       required: false,
       description: 'Name of the stack that needs to be created as csv filename.',
     }),
+    'stack-api-key': flags.string({
+      char: 'k',
+      multiple: false,
+      required: false,
+      description: 'Api key of the stack',
```
antonyagustine, line 40: Can we keep the description text the same as in another module like export/import?

netrajpatel (approved):

PR #846, opened 2023-06-19
Recursive backup folders created in import

aman19K:
Ticket:- https://contentstack.atlassian.net/browse/CS-34038


packages/contentstack-import/src/app.js
```diff
@@ -5,13 +5,15 @@
...
-const ncp = require('ncp');
+const os = require('os');
 const path = require('path');
 const chalk = require('chalk');
+const { copy, copySync, removeSync } = require('fs-extra');
 const util = require('./lib/util/index');
 const login = require('./lib/util/login');
 const { addlogs } = require('./lib/util/log');
 const { managementSDKClient, isAuthenticated } = require('@contentstack/cli-utilities');
+const { camelCase } = require('lodash')
```
netrajpatel, line 16: @aman19K you are using the lodash package in import but this package is not installed in the package.json file, this will lead to a package not found error while deployment.

netrajpatel (changes requested):

PR #847, opened 2023-06-20
Display changed content type properties in detail text mode

aman19K:
Ticket:- https://contentstack.atlassian.net/browse/CS-39275


shafeeqd959 (commented):
Have you checked merge command as well ? does it effect merge command ? pls check that as well

aman19K:
> @aman19K Have you checked merge command as well ? does it effect merge command ? pls check that as well


@shafeeqd959 I've checked merge command it's working as expected. We've simply changed detailed-text view.

PR #851, opened 2023-06-21
Oauth window updates

almeida-shrutika:
Ticket No.: CS-39326, CS-39327, CS-39333
No Version bump done


packages/contentstack-auth/test/unit/commands/tokens-add.test.ts
```diff
@@ -87,7 +76,7 @@ describe('Tokens Add Command', () => {
     inquireStub.restore();
   });
 
-  it('Add a token with invalid api key, should fail to add', async function () {
+  it.skip('Add a token with invalid api key, should fail to add', async function () {
```
antonyagustine, line 79: @almeida-shrutika Any specific reason to skip this test case.?

almeida-shrutika: @antonyagustine  Actually, the api-key validation implementation is removed from the code so I have skipped the test case

antonyagustine: If the current CLI version does not consider this specific case as valid, would it be possible to remove the test instead of skipping it?

almeida-shrutika: Sure will remove it

antonyagustine (approved):

antonyagustine (changes requested):
Could we review all the test cases that are currently being skipped? We can either remove the test case if it's deemed invalid or eliminate the skip condition altogether.

almeida-shrutika:
> Could we review all the test cases that are currently being skipped? We can either remove the test case if it's deemed invalid or eliminate the skip condition altogether.

@antonyagustine Done

antonyagustine (approved):

PR #855, opened 2023-06-26
bugfix/CS-39417 Changed the error_message to errorMessage in client.ts and added cliux.print to show the error

cs-raj:
Ticket: CS-39417
Version bump: 1.4.11


netrajpatel (approved):

netrajpatel (approved):

packages/contentstack-bootstrap/package.json
```diff
@@ -17,7 +17,7 @@
     "test:report": "nyc --reporter=lcov mocha \"test/**/*.test.js\""
   },
```
netrajpatel, line 18: @cs-raj Bootstrap package version is not updated

netrajpatel (changes requested):

packages/contentstack-bootstrap/package.json
cs-raj: Done

netrajpatel (approved):

PR #873, opened 2023-07-04
Tests/tokens unit tests

akashbhosale97:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=vfZ1olH)

Ticket  for delivery token test cases - https://contentstack.atlassian.net/browse/CS-39153
Ticket for delivery token test cases - https://contentstack.atlassian.net/browse/CS-39156

**Version bump not required**


packages/contentstack-auth/env.example
```diff
@@ -0,0 +1,8 @@
+BRANCH_ENABLED_API_KEY
+BRANCH_ENABLED_MGMT_TOKEN
+BRANCH_DISABLED_API_KEY
+BRANCH_DISABLED_MGMT_TOKEN
+BRANCH_ENABLED_DELIVERY_TOKEN
+BRANCH_DISABLED_DELIVERY_TOKEN
+BRANCH_ENABLED_ENVIRONMENT
+BRANCH_DISABLED_ENVIRONMENT
```
antonyagustine, line 8: @akashbhosale97 Looks like your test case utilising the original config. Instead of using the original config try setting the few keys in env and create your own config files after all test execution just clean up the config that is created for the test cases.

Env Keys that can create you new file. Check the config handler for more details.
```
CONFIG_NAME
ENC_CONFIG_NAME
```

For the cleanup script please refer [this](https://github.com/contentstack/cli/blob/tests/tokens-unit-tests/packages/contentstack-export/test/integration/clean-up.test.js#L29-L60)

packages/contentstack-auth/test/unit/commands/tokens-add.test.ts
```diff
@@ -100,3 +104,171 @@ describe('Tokens Add Command', () => {
...
+        process.env.BRANCH_DISABLED_DELIVERY_TOKEN!,
+        '-e',
+        process.env.BRANCH_DISABLED_ENVIRONMENT!,
+        '--branch',
+        branch,
+      ]);
+      assert.calledOnce(errorStub);
+    });
+  });
+});
```
antonyagustine, line 274: As discussed,

- please remove the deprecated flags and use the new flags and do not use any short flags.
- Check all test case execution times and verify if any test case taking longer time than expected if so increase the timeout.

antonyagustine (approved):

netrajpatel (approved):

PR #874, opened 2023-07-04
Rewrite release1

shafeeqd959:


packages/contentstack-export/src/types/default-config.ts
```diff
@@ -0,0 +1,168 @@
...
+    webhooks: string;
+    stacks: string;
+  };
+  preserveStackVersion: boolean;
+  fetchConcurrency: number;
+  writeConcurrency: number;
+  developerHubBaseUrl: string;
+  marketplaceAppEncryptionKey: string;
+  useNewModuleStructure: boolean;
+}
```
antonyagustine, line 168: Can we remove all the manually generated types and enable the declaration in its config to create these file types.?

packages/contentstack-export/src/utils/basic-login.ts
```diff
@@ -0,0 +1,58 @@
...
+  isAuthenticated,
+  cliux,
+  configHandler,
+  authHandler,
+} = require('@contentstack/cli-utilities');
+
+const login = async (config: ExternalConfig): Promise<any> => {
+  const client = await managementSDKClient(config);
+  if (config.email && config.password) {
+    const response = await client.login({ email: config.email, password: config.password }).catch(Promise.reject);
```
antonyagustine, line 23: `.catch(Promise.reject)` As per my understanding this reject part is actually doing nothing, correct me if I'm wrong.

PR #880, opened 2023-07-06
Feat/cs 39752

aman19K:
- [x] [CS-39752](https://contentstack.atlassian.net/browse/CS-39752)
- [x] [CS-39753](https://contentstack.atlassian.net/browse/CS-39753)


packages/contentstack-import/src/import/modules/extensions.ts
```diff
@@ -0,0 +1,124 @@
...
+  async importExtensions(): Promise<any> {
+    if (this.extensions === undefined || isEmpty(this.extensions)) {
+      log(this.importConfig, 'No Extensions Found', 'info');
+      return resolve();
+    }
+
+    const apiContent = values(this.extensions);
+
+    const onSuccess = ({ response, apiData: { uid, title } = { uid: null, title: '' } }: any) => {
+      this.extSuccess.push(response);
```
shafeeqd959, line 72: Is this code working, can you check this ?

shafeeqd959: this.extSuccess.push(response); this.extUidMapper[uid] = response.uid;,  this context would be from parent base class and it doesnt have these properties. did it work ?

shafeeqfalcon (approved):

PR #909, opened 2023-07-13
CS-38607 - add management token support

akashbhosale97:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=ALnOdzX)

-added management token support (alias)
-version bump to 1.4.13


packages/contentstack-seed/src/commands/cm/stacks/seed.ts
```diff
@@ -67,21 +74,25 @@ export default class SeedCommand extends Command {
...
   static aliases = ['cm:seed'];
 
   async run() {
     try {
       const { flags: seedFlags } = await this.parse(SeedCommand);
+      const managementTokenAlias = seedFlags.alias;
 
-      if (!isAuthenticated()) {
+      if (!isAuthenticated() && !managementTokenAlias) {
         this.error('You need to login, first. See: auth:login --help', {
```
netrajpatel, line 91: @akashbhosale97 Now if we are giving the option to provide a management token as well, the error message can also be changed right?

netrajpatel (changes requested):

packages/contentstack-seed/src/commands/cm/stacks/seed.ts
akashbhosale97: have changed the error message

netrajpatel (dismissed):

netrajpatel (approved):

netrajpatel (approved):

PR #944, opened 2023-07-25
Merge development to staging

aman19K:
- [x] We are removing the .github/workflows/secrets-scan.yml file. This action is sourced from an open-source GitHub repository, but it appears that the repository owner has deleted the repository.
cc @netrajpatel @antonyagustine


aman19K:
#945

PR #949, opened 2023-07-25
Update Management SDK version ^1.10.0 -> ~1.10.0

aman19K:
Update management SDK version ^1.10.0 -> ~1.10.0. For management SDK we need to use patch version instead of minor so that going forward SDK changes doesn't break CLI.


shafeeqd959 (approved):

antonyagustine (approved):
@aman19K Why do we have the `@contentstack/management` package used in many cli package.? Please add a tracking ticket to remove the dependency from other locations.

cc @netrajpatel

PR #952, opened 2023-07-26
Fix backup-dir flag in import command

aman19K:
CS-40364
- [x] create backup directory using config
- [x] enable newFileStructure export & import

### DIFF
--- a/packages/contentstack-import/src/config/index.ts
+++ b/packages/contentstack-import/src/config/index.ts
@@ -388,8 +388,7 @@ const config: DefaultConfig = {
   developerHubBaseUrl: '',
   marketplaceAppEncryptionKey: 'nF2ejRQcTv',
   getEncryptionKeyMaxRetry: 3,
-  useNewModuleStructure: true,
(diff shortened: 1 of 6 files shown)


packages/contentstack-import/src/config/index.ts
```diff
@@ -388,7 +388,8 @@ const config: DefaultConfig = {
   developerHubBaseUrl: '',
   marketplaceAppEncryptionKey: 'nF2ejRQcTv',
   getEncryptionKeyMaxRetry: 3,
-  useNewModuleStructure: false,
+  useNewModuleStructure: true,
+  createBackupDir: './temp'
```
antonyagustine, line 392: Do we need to keep the default value as "temp"? I assume we'll only accept this key from an external configuration, right?

aman19K: @antonyagustine Yeah, I missed that. Added it for testing. Removed

antonyagustine (approved):

shafeeqd959 (approved):

PR #977, opened 2023-08-09
CS-40469-Error while importing the global feilds refrences not removed

cs-raj:


packages/contentstack-import/src/utils/content-type-helper.ts
```diff
@@ -138,7 +147,7 @@ export const removeReferenceFields = async function (
     ) {
       flag.supressed = true;
       schema[i].reference_to = ['sys_assets'];
-      return true;
+      return true; // undo remove
```
abhinav-from-contentstack, line 150: @antonyagustine Removing line 150 is not part of Raj's solution (he had removed it on my suggestion). Adding `return true` again for now. This will be handled later, while addressing issue 40844

antonyagustine: @cs-raj and @abhinav-from-contentstack Can we have the above changes also in the same PR if it's just removing return statements only.?

aman19K: > @cs-raj and @abhinav-from-contentstack Can we have the above changes also in the same PR if it's just removing return statements only.?

@antonyagustine , @cs-raj  and @abhinav-from-contentstack  Yes, I think that return is unnecessary here, and by removing it, the prime demo stack issue also resolved.

abhinav-from-contentstack: Sure @antonyagustine

abhinav-from-contentstack: I've added those changes, please have a look @antonyagustine, @cs-raj

aman19K (dismissed):

abhinav-from-contentstack (dismissed):

abhinav-from-contentstack:
The changes from this branch have been merged directly to branch release/12171. Closing this PR.

PR #993, opened 2023-08-22
Release/12171

abhinav-from-contentstack:
- bumped cli version to 1.8.1
- bumped @contentstack/cli-cm-import version to 1.8.1
- added fixes for global field reference issue and RTE issue


packages/contentstack-import/src/utils/content-type-helper.ts
```diff
@@ -90,7 +90,7 @@ export const removeReferenceFields = async function (
       for (var block in schema[i].blocks) {
         await removeReferenceFields(schema[i].blocks[block].schema, flag, stackAPIClient);
       }
-    } else if (schema[i].data_type === 'reference' || schema[i].reference_to) {
+    } else if (schema[i].data_type === 'reference') {
```
antonyagustine, line 93: @shafeeqd959  / @aman19K  Could someone verify here whether removing this condition will have no impact on functionality?

While the previous code didn't have it, the rewrite shows some changes. Seeking clarity before approving the PR to avoid new issues while addressing the current one.

cc @netrajpatel @abhinav-from-contentstack @cs-raj

packages/contentstack-import/src/utils/content-type-helper.ts
```diff
@@ -138,7 +146,6 @@ export const removeReferenceFields = async function (
     ) {
       flag.supressed = true;
       schema[i].reference_to = ['sys_assets'];
-      return true;
```
antonyagustine, line 141: @shafeeqd959  / @aman19K  Could you confirm this change as well.?

cc @netrajpatel @abhinav-from-contentstack @cs-raj

cs-raj (dismissed):

packages/contentstack-import/src/utils/content-type-helper.ts
abhinav-from-contentstack: In `src/import/modules-js/global-field.js` file, removeReferenceFields sets value for `flag.supressed` property and based on the flag value we decide if global-field needs to be pushed to _globalField_pending.
https://github.com/contentstack/cli/blob/c005a1343aad6a16d041dc68327e52f1ee6df5e6/packages/contentstack-import/src/import/modules-js/global-fields.js#L66C10-L71C12

For the corresponding ts file, the dependence on `flag.supressed` has been removed. And the value returned by removeReferenceFields is being used to determine if global field needs to be pushed to pending array.
https://github.com/contentstack/cli/blob/c005a1343aad6a16d041dc68327e52f1ee6df5e6/packages/contentstack-import/src/import/modules/global-fields.ts#L119-L122

I think, this issue can be solved by re-introducing the flag in ts file again
https://github.com/contentstack/cli/blob/c005a1343aad6a16d041dc68327e52f1ee6df5e6/packages/contentstack-import/src/import/modules-js/global-fields.js#L63

cc: @shafeeqd959 @antonyagustine @aman19K

packages/contentstack-import/src/utils/content-type-helper.ts
abhinav-from-contentstack: I think this condition may have been added as a check if `reference_to` field exists in the object, as `reference_to` field is being used in a loop.
https://github.com/contentstack/cli/blob/c005a1343aad6a16d041dc68327e52f1ee6df5e6/packages/contentstack-import/src/utils/content-type-helper.ts#L93-L106

The problem is that objects that have `data_type: 'reference'` and `data_type: 'json'` ( in case of Json Rte ), both have `reference_to` fields.

As the condition on line 93 is an OR condition, objects with `data_type: 'json'` also get passed. And that should not happen, as this block is specifically written to handle `data_type: 'reference'` objects.

Although I've never seen an object with `data_type: reference` that doesn't have a `reference_to` array, if we are skeptical then maybe replacing `||` with `&&` will also work... atleast in the context of the Prime Demo issue.

@cs-raj may check if this also works for his fix.
Let me know what you guys think.
cc : @aman19K @antonyagustine @shafeeqd959

shafeeqd959 (approved):

packages/contentstack-import/src/utils/content-type-helper.ts
antonyagustine: @shafeeqd959  Kindly validate the accuracy of the identified observations and the proposed suggestion.

shafeeqd959: @abhinav-from-contentstack why do we need to remove the return statement here, if we remove it that will break the rewrite/ts version of global field import. Please let me know

aman19K: @shafeeqd959 If there are many fields in the global field, such as json_rte, block, and group, and we reference CT in each field, json_rte will return right after json_rte and won't remove references to block and group. It appears that we are presuming a sequence of fields just from looking at the code. Correct me if I'm wrong.

cc: @abhinav-from-contentstack @antonyagustine

packages/contentstack-import/src/utils/content-type-helper.ts
cs-raj: @abhinav-from-contentstack I tried the above it imported the global fields without issue from the selinc stack.
cc: @aman19K @antonyagustine @shafeeqd959

packages/contentstack-import/src/import/modules-js/entries.js
```diff
@@ -1218,13 +1218,17 @@ module.exports = class ImportEntries {
           break;
         }
         case 'json': {
+          const structuredPTag = '{"type":"p","attrs":{},"children":[{"text":""}]}';
```
shafeeqd959, line 1221: @abhinav-from-contentstack , can we make this same changes in the ts module as well. In ts we have this json rte handler in utilities

packages/contentstack-import/src/utils/content-type-helper.ts
abhinav-from-contentstack: This is correct, I had a discussion with @shafeeqd959 and we have come to the conclusion that re-introducing `flag.supressed` in the `ts` version is the way to fix this issue for now.

packages/contentstack-import/src/import/modules-js/entries.js
abhinav-from-contentstack: https://github.com/contentstack/cli/blob/63413c768ae87a37a8b2eab1bdc9e3a24b8550a1/packages/contentstack-import/src/utils/entries-helper.ts#L376

antonyagustine (approved):

shafeeqd959 (approved):

aman19K (approved):

PR #999, opened 2023-08-23
Bugfix/CS-40968 - added setWorkflow column to include the object as string, modified message in chooseContentType to select contentype

cs-raj:
version bump: 1.4.1


cs-raj:
The changes mentioned here are handled in #995

PR #1026, opened 2023-09-06
Audit plugin phase 1

antonyagustine:
**CS-41076**


packages/contentstack/package.json
```diff
@@ -37,11 +38,10 @@
     "@contentstack/cli-migration": "~1.3.12",
     "@contentstack/cli-utilities": "~1.5.2",
     "@contentstack/management": "~1.10.0",
-    "@contentstack/cli-cm-branches": "~1.0.12",
```
shafeeqd959, line 40: can we add this back ?

packages/contentstack-audit/bin/dev
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env node
```
shafeeqd959, line 1: can we use the latest method for the dev ?

packages/contentstack-audit/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
shafeeqd959, line 1: can we use prettier config instead of editor ?

packages/contentstack-audit/bin/dev
antonyagustine: What do you mean by `latest method for the dev`? Can you rephrase it.?

packages/contentstack-audit/.editorconfig
antonyagustine: Sure, I can remove it and commit. However, I need not push one for prettier we have it in our CLI global config itself.

packages/contentstack/package.json
antonyagustine: Sure

packages/contentstack-audit/.editorconfig
shafeeqd959: that works

packages/contentstack-audit/bin/dev
shafeeqd959: #!/usr/bin/env node
require('dotenv').config({ path: './.env' });
(async () => {
  const { execute } = require('@contentstack/cli-utilities');
  await execute({ type: 'cjs', development: true, dir: __dirname });
})();

we could use this instead

netrajpatel (dismissed):

aman19K (approved):

PR #1031, opened 2023-09-11
CS-41350-added .DS_Store to skipfile and version bump

cs-raj:
Version Bump: 1.8.4
Ticket: CS-41350
The file was already added to the .ts files for entries


packages/contentstack-import/src/import/modules-js/entries.js
```diff
@@ -38,7 +38,7 @@ module.exports = class ImportEntries {
   ctPath;
   lPath;
   importConcurrency;
-  skipFiles = ['__master.json', '__priority.json', 'schema.json'];
+  skipFiles = ['__master.json', '__priority.json', 'schema.json','.DS_Store'];
```
netrajpatel, line 41: @cs-raj Are there any similar TS changes required in the TS version of entries.js?

netrajpatel (changes requested):

packages/contentstack-import/src/import/modules-js/entries.js
cs-raj: No @netrajpatel. This is already handled there

shafeeqd959 (approved):

aman19K (approved):

packages/contentstack-import/src/import/modules-js/entries.js
cs-raj: Correction from my end: I connected with Shafeqq and Learnt that even though we have skipFiles in content types. In both entries and content type we are reading from schema.json due to which the issue of .DS_Store won't arise

shafeeqd959 (approved):

netrajpatel (approved):

PR #1032, opened 2023-09-11
Extend "Export-to-CSV" support to include Taxonomies

aman19K:


packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
```diff
@@ -59,6 +60,9 @@ class ExportToCsvCommand extends Command {
       multiple: false,
       required: false,
     }),
+    'taxonomy-uid': flags.string({
+      description: 'Provide taxonomy UID for which terms need to be exported',
```
netrajpatel, line 64: @aman19K Can we use inclusive validation over the action?
If the action is taxonomies taxonomy-uid should be required.

netrajpatel (changes requested):

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
aman19K: @netrajpatel I've handled using prompt. If action is taxonomies and taxonomy-uid isn't provided then handling it thru prompt. https://github.com/contentstack/cli/pull/1032/files#diff-5b7cc0b05cb3c3355d3ae284fb8748b62364f914ecdce0c0fc846e3b103167efR226-R230

aman19K: @netrajpatel Removed above prompt as taxonomy-uid is a optional flag.

netrajpatel (approved):

PR #1047, opened 2023-09-20
Bugfix/CS-41110 -  Added the path of contentstack folder for the logs to be created

cs-raj:
Version Bump: 1.5.3


packages/contentstack-auth/src/commands/auth/base-command.ts
antonyagustine, line 1: Can this file moved root folder of the src or out side of the command folder.?

packages/contentstack-config/src/commands/config/base-command.ts
antonyagustine, line 1: Can you remove this as I can see another file which is already in the src directory.?

packages/contentstack-config/src/base-command.ts

packages/contentstack/src/commands/cm/stacks/base-command.ts
antonyagustine, line 1: As I couldn’t see this base command was never used. Can we remove this class if it's not used anywhere.?

antonyagustine (changes requested):

packages/contentstack-auth/src/commands/auth/base-command.ts
cs-raj: Done @antonyagustine

packages/contentstack/src/commands/cm/stacks/base-command.ts
cs-raj: Done @antonyagustine

packages/contentstack-config/src/commands/config/base-command.ts
cs-raj: Done @antonyagustine

antonyagustine (approved):

netrajpatel (approved):

PR #1049, opened 2023-09-21
fix: moved axios to dev dependencies

vkalta:
This PR resolves CS-41235.


packages/contentstack-seed/package.json
```diff
@@ -8,7 +8,6 @@
     "@contentstack/cli-cm-import": "~1.8.4",
```
netrajpatel, line 8: @vkalta Can you please bump the package version?

netrajpatel (changes requested):

netrajpatel:
@vkalta Please resolve the conflict.

packages/contentstack-seed/package.json
netrajpatel: @vkalta Package lock is not updated with the new version bump, can you please generate a new package-lock and push it?

packages/contentstack-seed/package.json
```diff
@@ -1,14 +1,13 @@
 {
   "name": "@contentstack/cli-cm-seed",
   "description": "create a Stack from existing content types, entries, assets, etc.",
-  "version": "1.5.0",
+  "version": "1.5.1",
```
netrajpatel, line 4: @vkalta package-lock is not updated with new bumped version.
Can you push the updated package-lock file?

netrajpatel (changes requested):

netrajpatel (approved):

netrajpatel (approved):

PR #1052, opened 2023-09-21
Add Taxonomy module to CLI Export Plugin

aman19K:
**CS-41090**

- [x] Add taxonomy support in CLI export

### DIFF
--- a/packages/contentstack-export/src/config/index.ts
+++ b/packages/contentstack-export/src/config/index.ts
@@ -25,6 +25,7 @@ const config: DefaultConfig = {
       'environments',
       'extensions',
       'webhooks',
+      'taxonomies',
       'global-fields',
       'content-types',
       'custom-roles',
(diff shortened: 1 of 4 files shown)


packages/contentstack-export/src/config/index.ts
```diff
@@ -25,6 +25,7 @@ const config: DefaultConfig = {
       'environments',
       'extensions',
       'webhooks',
+      'taxonomies',
```
netrajpatel, line 28: @aman19K I hope this is considering the dependency of the plugin hierarchy

netrajpatel (approved):

packages/contentstack-export/src/config/index.ts
aman19K: Yes @netrajpatel

antonyagustine (approved):

PR #1062, opened 2023-09-26
Add Taxonomy module to CLI Import Plugin

aman19K:
CS-41385
- [x] Import taxonomies
- [x] Import terms


packages/contentstack-import/src/import/modules/base-class.ts
```diff
@@ -381,6 +384,19 @@ export default abstract class BaseClass {
...
+          .post(additionalInfo.url, { taxonomy: apiData })
+          .then(onSuccess)
+          .catch(onReject);
+      case 'create-terms':
+        const url = `${additionalInfo.baseUrl}/${apiData.taxonomy_uid}/terms`;
+        return new HttpClient()
+          .headers(additionalInfo.headers)
+          .post(url, { term: apiData })
+          .then(onSuccess)
+          .catch(onReject);
```
antonyagustine, line 399: Can this `HttpClient` instance be created globally instead of being created every time the case is triggered, considering its usage in multiple locations?

antonyagustine (approved):

PR #1079, opened 2023-10-05
Fix/cs 41654

abhinav-from-contentstack:
- added error handling in create-update entries script
- updated branches plugin to v1.0.15


abhinav-from-contentstack:
Hi @aman1908, I've removed all the console statements. So, the errors will get logged as they did before.
But there are not many changes other than that. Do we need this PR?

aman19K:
> Hi @aman1908, I've removed all the console statements. So, the errors will get logged as they did before. But there are not many changes other than that. Do we need this PR?

@abhinav-from-contentstack we can close this PR.

aman19K (approved):

antonyagustine (approved):

PR #1107, opened 2023-10-17
Feat/import overwrite base

shafeeqd959:


netrajpatel (dismissed):

packages/contentstack-import/src/import/modules/entries.ts
```diff
@@ -402,6 +449,135 @@ export default class EntriesImport extends BaseClass {
...
+          .entry()
+          .query({ query: { title: entry.title, locale } })
+          .findOne()
+          .catch((error: Error) => {
+            apiParams.reject({
+              error,
+              apiData: entry,
+            });
+            reject(true);
+          })) || {};
```
antonyagustine, line 547: [MINOR] Can we take this block of code out of Promise? so that we do need not to use async inside the promise which will throw a lint issue.

packages/contentstack-import/src/import/modules/extensions.ts
```diff
@@ -109,19 +125,88 @@ export default class ImportExtensions extends BaseClass {
...
+        .extension()
+        .query({ query: { title: extension.title } })
+        .findOne()
+        .catch((error) => {
+          apiParams.reject({
+            error,
+            apiData: extension,
+          });
+          reject(true);
+        });
```
antonyagustine, line 178: [MINOR] Can we take this block of code out of Promise? so that we do need not to use async inside the promise which will throw a lint issue.

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -140,4 +159,51 @@ export default class ImportGlobalFields extends BaseClass {
...
+    };
+
+    await this.makeConcurrentCall(
+      {
+        apiContent: this.existingGFs,
+        processName: 'Replace global fields',
+        apiParams: {
+          serializeData: this.serializeReplaceGFs.bind(this),
+          reject: onReject.bind(this),
+          resolve: onSuccess.bind(this),
```
antonyagustine, line 184: [MINOR] Need not bind the current object into it. However, the current code will still work the way we expect.

packages/contentstack-import/src/import/modules/locales.ts
```diff
@@ -137,9 +137,14 @@ export default class ImportLocales extends BaseClass {
...
-      log(this.importConfig, `Language '${code}' failed to import`, 'error');
-      log(this.importConfig, formatError(error), 'error');
-      this.failedLocales.push({ uid, code });
+      if (error?.errorCode === 247) {
+        log(this.importConfig, formatError(error), 'info');
+        this.failedLocales.push({ uid, code });
+      } else {
+        log(this.importConfig, `Language '${code}' failed to import`, 'error');
+        log(this.importConfig, formatError(error), 'error');
+        this.failedLocales.push({ uid, code });
```
antonyagustine, line 146: ` this.failedLocales.push({ uid, code });` This line can be out of both if and else condition

packages/contentstack-import/src/import/modules/extensions.ts
```diff
@@ -109,19 +125,88 @@ export default class ImportExtensions extends BaseClass {
...
+      log(this.importConfig, error, 'error');
+    };
+
+    await this.makeConcurrentCall(
+      {
+        apiContent: this.existingExtensions,
+        processName: 'Replace extensions',
+        apiParams: {
+          reject: onReject.bind(this),
+          resolve: onSuccess.bind(this),
```
antonyagustine, line 148: [MINOR] Need not bind the current object into it. However, the current code will still work the way we expect.

abhinav-from-contentstack (dismissed):

packages/contentstack-import/src/import/modules/entries.ts
shafeeqd959: @antonyagustine , this particular code is disabled using a overwrite flag, changing structure would need more testing. will consider this point on the phase 2 when entries will be enabled. For now not getting any lint issues

packages/contentstack-import/src/import/modules/extensions.ts
shafeeqd959: @antonyagustine , its not giving any lint issues as such, will consider this point on phase 2 as it takes more time to develop and test

packages/contentstack-import/src/import/modules/extensions.ts
shafeeqd959: @antonyagustine , will take this on phase2, since this is done for all modules, we need to update everywhere and test again

packages/contentstack-import/src/import/modules/global-fields.ts
shafeeqd959: @antonyagustine will take this on phase2, since this is done for all modules, we need to update everywhere and test again

packages/contentstack-import/src/import/modules/locales.ts
```diff
@@ -137,8 +137,12 @@ export default class ImportLocales extends BaseClass {
       fsUtil.writeFile(this.langUidMapperPath, this.langUidMapper);
     };
     const onReject = ({ error, apiData: { uid, code } = undefined }: any) => {
-      log(this.importConfig, `Language '${code}' failed to import`, 'error');
-      log(this.importConfig, formatError(error), 'error');
+      if (error?.errorCode === 247) {
+        log(this.importConfig, formatError(error), 'info');
+      } else {
+        log(this.importConfig, `Language '${code}' failed to import`, 'error');
+        log(this.importConfig, formatError(error), 'error');
```
antonyagustine, line 144: @shafeeqd959 Even this line looks like same in both the if and else blocks except the print type (info/error). BTW, Why is the error printed as info.?

PR #1108, opened 2023-10-18
Feat/CS-41971 - Using teams api call to export all the teams data into csv format 

cs-raj:
Using teams api call to export all the teams data into csv format and adding the team-uid flag to export the user data for specific teams. 

- Version bump: no
- This PR also contains the changes for the ticket: CS-41982


netrajpatel:
> Version bump: no

@cs-raj Why the version bump not required or not done for this PR?

cs-raj:
> > Version bump: no
> 
> @cs-raj Why the version bump not required or not done for this PR?

When I will be raising a PR from [feat/CS-40615-teams] to development I will then bump the version collectively

abhinav-from-contentstack (approved):

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
```diff
@@ -59,6 +59,11 @@ class ExportToCsvCommand extends Command {
       multiple: false,
       required: false,
     }),
+    "team-uid": flags.string({
+      description: 'Uid of the team whose user data and stack roles are required',
+      multiple: false,
```
aman19K, line 64: @cs-raj, If team-uid is an optional flag, then required and multiple don't need to be included.

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -6,10 +6,9 @@ const fastcsv = require('fast-csv');
 const inquirer = require('inquirer');
 const debug = require('debug')('export-to-csv');
 const checkboxPlus = require('inquirer-checkbox-plus-prompt');
-
 const config = require('./config.js');
-const { cliux, configHandler } = require('@contentstack/cli-utilities');
-
+const { cliux, configHandler, HttpClient } = require('@contentstack/cli-utilities');
+const _ = require('lodash');
```
aman19K, line 11: @cs-raj Instead of loading the full library, we can retrieve a specific method from Lodash. like const find = require('lodash/find');

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -678,6 +683,174 @@ function wait(time) {
...
+}
+
+async function exportOrgTeams(managementAPIClient, org) {
+  let teamsObjectArray = [];
+  let skip = 0;
+  let limit = config.limit || 100;
+  do {
+    const data = await apiRequestHandler(org, { skip: skip, limit: limit, includeUserDetails: true });
+    skip += limit;
+    teamsObjectArray = [...teamsObjectArray, ...data?.teams];
```
aman19K, line 730: @cs-raj, Don't require spread operator here. We can use teamsObjectArray.push(...data.items) instead of spread operator.

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -678,6 +683,174 @@ function wait(time) {
...
+    if (skip > data?.count) break;
```
aman19K, line 731: @cs-raj, Can you validate this case with 100, 200 data count ? I think we can avoid one more iteration using skip >= data?.count.

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -678,6 +683,174 @@ function wait(time) {
...
+    'createdAt',
+    'createdBy',
+    'updatedAt',
+    'updatedBy',
+    '__v',
+    'createdByUserName',
+    'updatedByUserName',
+    'organizationUid',
+  ];
+  if (data?.length !== 0) {
```
aman19K, line 770: @cs-raj, I believe that if (data?.length) is sufficient to verify whether a statement is true or false. We don't need data?.length !== 0.

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -678,6 +683,174 @@ function wait(time) {
...
+        team.organizationRole = 'member';
+      } else {
+        team.organizationRole = 'admin';
+      }
+      team.Total_Members = team?.users?.length || 0;
+    });
+  } else {
+    cliux.print(`info: There are no teams in the given org`);
+  }
+  return data;
```
aman19K, line 788: @cs-raj, Please check the return statement and also verify the case when there is no team in the given org. I think return statement is missing in else condition.

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -678,6 +683,174 @@ function wait(time) {
...
+}
+
+async function exportTeams(managementAPIClient, organization, teamUid) {
+  const allTeamsData = await exportOrgTeams(managementAPIClient, organization);
+  if (allTeamsData.length === 0) {
+    this.log(`There are not teams in the organization named ${organization?.name}`);
```
aman19K, line 794: @cs-raj, Could you pls confirm this.log ? I think in this case it won't work as we're not inherting command class here.
And also instead of if (allTeamsData.length === 0) we can use if (!allTeamsData?.length).

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -678,6 +683,174 @@ function wait(time) {
...
+    
+    write(this, team.users, fileName, 'Team User details');
+  
+  }
+}
+
+async function getTeamsUserDetails(teamsObject) {
+  const allTeamUsers = [];
+  teamsObject.forEach((team) => {
+    if (team.users.length) {
```
aman19K, line 841: @cs-raj, here we can use optional chaining.
```suggestion
    if (team.users.length) {
```

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
cs-raj: Removed the fields

packages/contentstack-export-to-csv/src/util/index.js
cs-raj: Imported clone deep only

packages/contentstack-export-to-csv/src/util/index.js
cs-raj: Yes, used it like mentioned

packages/contentstack-export-to-csv/src/util/index.js
cs-raj: Yes it was sufficient to verify removed the redundant condition

packages/contentstack-export-to-csv/src/util/index.js
cs-raj: Yes added

packages/contentstack-export-to-csv/src/util/index.js
cs-raj: Yes this.log gave me the error when i tested with org having no teams therefore removed it and used console.log instead

packages/contentstack-export-to-csv/src/util/index.js
cs-raj: Validated @aman19K by changing the limits

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
```diff
@@ -59,6 +59,9 @@ class ExportToCsvCommand extends Command {
       multiple: false,
       required: false,
     }),
+    "team-uid": flags.string({
+      description: 'Uid of the team whose user data and stack roles are required'
+    })
```
antonyagustine, line 64: Have we reviewed this text with doc team.?

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
```diff
@@ -258,6 +262,32 @@ class ExportToCsvCommand extends Command {
...
           break;
         }
+        case config.exportTeams:
+        case 'teams': {
+          try{
+            if (!isAuthenticated()) {
+              this.error(config.CLI_EXPORT_CSV_LOGIN_FAILED, {
+                exit: 2,
+                suggestions: ['https://www.contentstack.com/docs/developers/cli/authentication/'],
+              });
```
antonyagustine, line 272: We could have added this auth check as common rather then duplicating every switch cases.

packages/contentstack-export-to-csv/src/util/config.js
```diff
@@ -1,10 +1,11 @@
...
-  exportUsers: "Export organization users' data to a .CSV file",
+  exportUsers: "Export organization user's data to a .CSV file",
+  exportTeams: "Export organization team's data to a .csv file",
   adminError: "Unable to export data. Make sure you're an admin or owner of this organization",
   organizationNameRegex: /\'/,
   CLI_EXPORT_CSV_LOGIN_FAILED: "You need to login to execute this command. See: auth:login --help",
-  CLI_EXPORT_CSV_ENTRIES_ERROR: "You need to either login or provide a management token to execute this command"
-
+  CLI_EXPORT_CSV_ENTRIES_ERROR: "You need to either login or provide a management token to execute this command",
+  CLI_EXPORT_CSV_API_FAILED: 'Something went wrong. Please try again!'
```
antonyagustine, line 10: Is this text reviewed by doc team?

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -6,10 +6,9 @@ const fastcsv = require('fast-csv');
 const inquirer = require('inquirer');
 const debug = require('debug')('export-to-csv');
 const checkboxPlus = require('inquirer-checkbox-plus-prompt');
-
 const config = require('./config.js');
-const { cliux, configHandler } = require('@contentstack/cli-utilities');
-
+const { cliux, configHandler, HttpClient } = require('@contentstack/cli-utilities');
+const {cloneDeep} = require('lodash');
```
antonyagustine: Can we do a specific import to improve the import cost.?

`const cloneDeep = require('lodash/cloneDeep');`

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -678,6 +683,174 @@ function wait(time) {
   });
 }
 
+function handleErrorMsg(err) {
+  if (err?.errorMessage) {
+    cliux.print(`Error: ${err.errorMessage}`, { color: 'red' });
+  } else if (err?.message) {
+    cliux.print(`Error: ${err.message}`, { color: 'red' });
+  } else {
```
antonyagustine, line 691: Your code going to works fine. Still can you try combain this if possible and that realy makes more readable for other developers?

`cliux.print("Error: ${err?.errorMessage ?? err?.message}", { color: 'red' });`

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -678,6 +683,174 @@ function wait(time) {
...
+    .then((roles) => {
+      roles.items.forEach((item) => {
+        if (item.name === 'member' || item.name === 'admin') {
+          roleMap.name = item.uid;
+        }
+      });
+    })
+    .catch((err) => {
+      handleErrorMsg(err);
+    });
```
antonyagustine, line 753: re-confirm with the respective ticket owner do we need to consider the pagination here.

aman19K (approved):

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
cs-raj: Not yet, I am preparing the sheet

packages/contentstack-export-to-csv/src/util/config.js
cs-raj: Yes this one is reviewed it is taken from taxonomy

packages/contentstack-export-to-csv/src/util/index.js
cs-raj: Yes used

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
cs-raj: Done

packages/contentstack-export-to-csv/src/util/index.js
cs-raj: Got it . modified it

packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -678,6 +683,168 @@ function wait(time) {
...
+        team.description = '';
+      }
+      if (team.organizationRole === roleMap['member']) {
+        team.organizationRole = 'member';
+      } else {
+        team.organizationRole = 'admin';
+      }
+      team.Total_Members = team?.users?.length || 0;
+    });
+    return data;
```
antonyagustine, line 779: Your code looks good and absolutely will do the job.

[MINOR, OPTIONAL] We can do a little bit of improvement on top of your logic as follows.

```
return data.map((team) => {
  // Lodash dependency: import omit from 'lodash/omit'
  team = omit(team, fieldToBeDeleted);

  team.organizationRole = team.organizationRole === roleMap["member"] ? "member" : "admin";

  if (!team.hasOwnProperty("description")) {
    team.description = "";
  }
  team.Total_Members = team?.users?.length || 0;

  return team;
});
```

antonyagustine (approved):

packages/contentstack-export-to-csv/src/util/index.js
cs-raj: Done

aman19K (approved):

PR #1109, opened 2023-10-18
Feat/CS-41982 - Exporting the teams data and added flag team-uid if flag is provided then specific team user data is written else all teams user data

cs-raj:


cs-raj:
The changes here are made in #1108

PR #1111, opened 2023-10-18
Taxonomy | Internal demo suggestions and sdk issue

aman19K:
- [x] Internal demo suggestions(CS-42076)
- [x] SDK taxonomies find issue (1.11.0 -> 1.12.0)

### DIFF
--- a/packages/contentstack-export-to-csv/src/util/index.js
+++ b/packages/contentstack-export-to-csv/src/util/index.js
@@ -702,7 +702,7 @@ async function getAllTaxonomies(payload, skip = 0, taxonomies = []) {
   const { items, count } = await taxonomySDKHandler(payload, skip);
   if (items) {
     skip += payload.limit;
-    taxonomies = [...taxonomies, ...items];
+    taxonomies.push(...items);
     if (skip >= count) {
       return taxonomies;
     } else {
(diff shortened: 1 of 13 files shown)


packages/contentstack-export-to-csv/src/util/index.js
```diff
@@ -699,12 +699,11 @@ function wait(time) {
...
   payload['type'] = 'taxonomies';
-  const response = await taxonomySDKHandler(payload, skip);
-  if (response) {
+  const { items, count } = await taxonomySDKHandler(payload, skip);
+  if (items) {
     skip += payload.limit;
-    //TODO - replace response.taxonomies with items
-    taxonomies = [...taxonomies, ...response.taxonomies];
-    if (skip >= response?.count) {
+    taxonomies = [...taxonomies, ...items];
```
antonyagustine, line 705: Is it necessary to spread both arrays.? The spread operator is equivalent to `for..of` loop [`O(n)`]. Instead, can we do `taxonomies.push(...items)`

antonyagustine (approved):

packages/contentstack-export-to-csv/src/util/index.js
aman19K: @antonyagustine sure, in this case push is better than spread operator. Done

shafeeqd959 (approved):

PR #1126, opened 2023-10-26
fix: added pathValidator utility method

vkalta:


aman19K (approved):

abhinav-from-contentstack:
There is an issue with the fix @vkalta. When a user enters absolute path, the validation is working incorrectly on file extensions as well. 
cc : @AniketDev7 @antonyagustine @netrajpatel

PR #1136, opened 2023-10-31
fix: incorrect date format in merge script

aman19K:
CS-42204
- [x] Utility function that generates a valid date and time format
- [x] version bumped
@contentstack/cli-utilities: ~1.5.5 -> ~1.5.6
@contentstack/cli-cm-branches: ~1.0.16 -> ~1.0.17


packages/contentstack-utilities/src/date-time.ts
```diff
@@ -0,0 +1,15 @@
...
+  ].join('');
+};
+
+export const formatTime = (date: Date) => {
+  return [
+    date.getHours().toString().padStart(2, '0'),
+    date.getMinutes().toString().padStart(2, '0'),
+    date.getSeconds().toString().padStart(2, '0'),
+  ].join('');
+};
```
antonyagustine, line 15: **General recommendations:**
- Use string concatenation for small strings.
- Use array join for large strings.

packages/contentstack-utilities/src/date-time.ts
```diff
@@ -0,0 +1,15 @@
+export const formatDate = (date: Date) => {
+  return [
+    date.getFullYear(),
```
shafeeqd959, line 3: Improve the file name form date-time to the actual purpose, like date-time-formatter

antonyagustine (approved):

packages/contentstack-utilities/src/date-time.ts
antonyagustine: The date-time itself is self-explanatory, implying that it can encompass various date-time-related utilities without the necessity of being strictly formatted dates or time.

PR #1138, opened 2023-11-01
Bugfix/CS-41007 -  added validation for management token in utilties 

cs-raj:
Version bump:
"@contentstack/cli-audit": "1.2.1",
"@contentstack/cli-auth": "1.3.16",
"@contentstack/cli-cm-bootstrap": "1.6.1",
"@contentstack/cli-cm-branches": "1.0.16",
"@contentstack/cli-cm-bulk-publish": "1.3.14",
"@contentstack/cli-cm-clone": "1.6.1",
"@contentstack/cli-cm-export": "1.9.3",
"@contentstack/cli-cm-export-to-csv": "1.4.5",
"@contentstack/cli-cm-import": "1.10.1",
"@contentstack/cli-cm-migrate-rte": "1.4.14",
"@contentstack/cli-cm-seed": "1.6.2",
"@contentstack/cli-command": "1.2.15",
"@contentstack/cli-config": "1.4.14",
"@contentstack/cli-launch": "1.0.14",
"@contentstack/cli-migration": "1.3.15",
"@contentstack/cli-utilities": "1.5.5",
"@contentstack/cli": "1.10.1"


packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
```diff
@@ -109,6 +110,11 @@ class ExportToCsvCommand extends Command {
                 apiKey: listOfTokens[managementTokenAlias].apiKey,
                 token: listOfTokens[managementTokenAlias].token,
               };
+              const validObject = await isManagementTokenValid(stackAPIKey,listOfTokens[managementTokenAlias].token);
```
shafeeqd959, line 113: can you name the "validObject" better

packages/contentstack-utilities/src/helpers.ts
```diff
@@ -8,3 +9,20 @@ export const doesBranchExist = async (stack, branchName) => {
       return error;
     });
 };
+
+export const isManagementTokenValid = async (stackAPIKey, managementToken) => {
+  const httpClient = new HttpClient({ headers: { api_key: stackAPIKey, authorization: managementToken } });
+  try {
+    const response = (await httpClient.get(`${configHandler.get('region').cma}/v3/environments?limit=1`))?.data;
```
shafeeqd959, line 16: "${configHandler.get('region').cma" consider a case where region is set and the user went to add the token, can you check the case ? handle the case if it is not handled the case already.

shafeeqd959 (approved):

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
cs-raj: yes changed its name

packages/contentstack-bootstrap/package.json
```diff
@@ -17,7 +17,7 @@
     "test:report": "nyc --reporter=lcov mocha \"test/**/*.test.js\""
   },
   "dependencies": {
-    "@contentstack/cli-cm-seed": "~1.6.1",
+    "@contentstack/cli-cm-seed": "~1.6.2",
```
shafeeqd959, line 20: Why did we bump the version is seed ? why didnt we bump it in export-to-csv, Pls update the contentstack package json as well with the version bump

cs-raj: seed was using utilities that's why bumped it's version also. Contentstacks version was also bumped I updated the branch with development changes that's why its not showing now

shafeeqd959: what about "export-to-csv", is it bumped already

abhinav-from-contentstack (approved):

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
```diff
@@ -119,6 +120,11 @@ class ExportToCsvCommand extends Command {
                 apiKey: listOfTokens[managementTokenAlias].apiKey,
                 token: listOfTokens[managementTokenAlias].token,
               };
+              const checkManagementTokenValidity = await isManagementTokenValid(stackAPIKey,listOfTokens[managementTokenAlias].token);
+              if(!checkManagementTokenValidity?.valid) {
+                cliux.print(`error: Management token or stack API key is invalid. ${checkManagementTokenValidity?.message||""}`, {color: "red"});
+                process.exit(1);
```
antonyagustine, line 126: @cs-raj Try to avoid exiting the process in the middle of the code. Doing so will prevent you from being able to cover those lines of code when writing test cases.

packages/contentstack-utilities/src/helpers.ts
```diff
@@ -8,3 +9,21 @@ export const doesBranchExist = async (stack, branchName) => {
...
+    if (response?.environments) {
+      return { valid: true, message: `valid token and stack api key` }
+    } else if(response?.error_code) {
+      return { valid: false, message: response.error_message };
+    } else {
+      throw typeof response === "string"? response : "";
+    }
+  } catch (error) {
+    cliux.print(`Failed to check the validity of the Management token. ${error}`, {color:"red"});
+    process.exit(1)
```
antonyagustine, line 27: Same comment mentioned above applies here as well. Did we covered the unit test cases for these lines?

antonyagustine (approved):

abhinav-from-contentstack (dismissed):

packages/contentstack-utilities/src/helpers.ts
cs-raj: Removed the process.exit

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
cs-raj: Removed the process.exit

antonyagustine (approved):

shafeeqd959 (approved):

packages/contentstack-utilities/src/helpers.ts
```diff
@@ -8,3 +9,20 @@ export const doesBranchExist = async (stack, branchName) => {
...
+  const httpClient = new HttpClient({ headers: { api_key: stackAPIKey, authorization: managementToken } });
+  try {
+    const response = (await httpClient.get(`${configHandler.get('region').cma}/v3/environments?limit=1`))?.data;
+
+    if (response?.environments) {
+      return { valid: true, message: `valid token and stack api key` }
+    } else if(response?.error_code) {
+      return { valid: false, message: response.error_message };
+    } else {
+      throw typeof response === "string"? response : "";
```
shafeeqd959, line 23: @cs-raj if response is not a string, it throws an empty string, can we keep some default message here so that user wont be confused

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
```diff
@@ -117,6 +118,11 @@ class ExportToCsvCommand extends Command {
             let contentTypes = [];
 
             if (managementTokenAlias) {
+              const listOfTokens = configHandler.get('tokens');
+              const checkManagementTokenValidity = await isManagementTokenValid(stackAPIKey,listOfTokens[managementTokenAlias].token);
+              if(!checkManagementTokenValidity?.valid) {
```
shafeeqd959, line 123: @cs-raj , its better to destructure valid and message values from "checkManagementTokenValidity" object and have a default values for it. Instead of using ? multiple places  in this code

packages/contentstack-utilities/src/helpers.ts
cs-raj: @shafeeqd959 when the response would not be string we are having default message as  **message:`Failed to check the validity of the Management token.**.

packages/contentstack-export-to-csv/src/commands/cm/export-to-csv.js
cs-raj: I would need that as an object as when we fail to validate the management token validity we are not passing the valid property

PR #1150, opened 2023-11-08
feat: migration script to read a CSV file and import taxonomy

aman19K:
CS-42278
- [x] Extend "Migration plugin" to support import taxonomy utility
- [x] version already bumped

### DIFF
--- a/packages/contentstack-migration/examples/taxonomies/import-taxonomies.js
+++ b/packages/contentstack-migration/examples/taxonomies/import-taxonomies.js
@@ -2,7 +2,7 @@ const fs = require('fs');
 const fastcsv = require('fast-csv');
 const find = require('lodash/find');
 
-module.exports = async ({ migration, stackSDKInstance, managementAPIClient, config }) => {
+module.exports = ({ migration, stackSDKInstance, managementAPIClient, config }) => {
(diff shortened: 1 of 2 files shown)


shafeeqd959 (approved):

cs-raj (approved):

abhinav-from-contentstack (approved):

packages/contentstack-migration/examples/taxonomies/import-taxonomies.js
```diff
@@ -0,0 +1,150 @@
+const fs = require('fs');
+const fastcsv = require('fast-csv');
+const find = require('lodash/find');
+
+module.exports = async ({ migration, stackSDKInstance, managementAPIClient, config }) => {
```
antonyagustine, line 5: Is this function expected to be asynchronous? As I don't see any await function!

antonyagustine (approved):

packages/contentstack-migration/examples/taxonomies/import-taxonomies.js
aman19K: @antonyagustine removed. Correct, we don't need async here; I kept it async because I was following an older script file.

PR #1200, opened 2023-12-04
fix: window os absolute path issue in migration command

aman19K:
- [x] window os absolute path issue in migration command (CS-42853)
- [x] version bump @contentstack/cli-migration': ~1.4.1 -> ~1.4.2


cs-raj (approved):

packages/contentstack-migration/src/commands/cm/stacks/migration.js
```diff
@@ -75,8 +75,10 @@ class MigrationCommand extends Command {
 
     if (Array.isArray(config) && config.length > 0) {
       let configObj = config.reduce((a, v) => {
-        let objArr = v.split(':');
-        return { ...a, [objArr[0]]: objArr[1] };
+        //NOTE: Temp code to handle only one spilt(Window absolute path issue).Need to replace with hardcoded config key
+        let [key, ...value] = v.split(':');
+        value = value?.length > 1 ? value?.join(':') : value?.join( );
+        return { ...a, [key]: value };
```
antonyagustine, line 81: Can we have a detailed tracking ticket if the current fix is considered a temp code as mentioned in the note/comment.?

antonyagustine (approved):

packages/contentstack-migration/src/commands/cm/stacks/migration.js
aman19K: Yes, @antonyagustine CS-42979

PR #1202, opened 2023-12-05
refactor: marketplace apps HTTP adapter calls to SDK call migration on import command [CS-42957]

antonyagustine:
**CS-42957**
- [x] refactor: marketplace apps HTTP adapter calls to SDK call migration on import command.

NOTE: As the new marketplace app SDK release is required, so version bump will be done later.


packages/contentstack-import/src/import/modules/marketplace-apps.ts
```diff
@@ -373,15 +422,16 @@ export default class ImportMarketplaceApps {
...
+      );
 
       if (installation.installation_uid) {
-        const appManifestName = app?.manifest?.name;
-        const appName = this.appNameMapping[appManifestName] ? this.appNameMapping[appManifestName] : appManifestName;
-
+        const appName = this.appNameMapping[app.manifest.name] ?? app.manifest.name;
         log(this.importConfig, `${appName} app installed successfully.!`, 'success');
-        await makeRedirectUrlCall(installation, appManifestName, this.importConfig);
+        await makeRedirectUrlCall(installation, app.manifest.name, this.importConfig);
```
aman19K, line 434: @antonyagustine , The app name has been assigned to the appName variable at Line 432. Thus, rather of using app.manifest.name in Line 434, should we use this?

aman19K (approved):

shafeeqd959:
@antonyagustine please do create a separate ticket to bump the versions of the plugins which have updates. Also please have a plan, if in any case markeplace sdk release is delayed.

CC: @netrajpatel

PR #1205, opened 2023-12-07
Fix: Entry import fails if JSON RTE has a reference inside the table

antonyagustine:
- [x] Fix: Extension UID mapper fix.
- [x] Fix: Entry import fails if JSON RTE has a reference inside the table.


packages/contentstack-import/src/import/modules/extensions.ts
```diff
@@ -83,7 +83,12 @@ export default class ImportExtensions extends BaseClass {
       return;
     }
 
-    const apiContent = values(this.extensions);
+    const apiContent = [];
+
+    for (const [uid, content] of Object.entries(this.extensions)) {
+      (content as Record<string, any>).uid = uid;
+      apiContent.push(content);
+    }
```
aman19K: Hey @antonyagustine , This code we don't require here.  While rewriting the extension export we are storing the data in different format .

cs-raj (approved):

aman19K (dismissed):

aman19K (approved):

PR #1221, opened 2023-12-20
Fix: Audit command schema not found fix

antonyagustine:
**CS-43193**

- [x] Audit fix command schema not found fixes.
- [x] Folder clean script fix on command completion.
- [x] Modular blocks issue fixed in content type.


packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -623,19 +630,21 @@ export default class Entries {
         return this.fixJsonRteMissingReferences([...tree, { index, type: child?.type, uid: child?.uid }], field, child);
       }) as EntryJsonRTEFieldDataType[];
     } else {
-      entry.children = entry.children
-        .map((child) => {
-          const refExist = this.jsonRefCheck(tree, field, child);
+      if (entry?.children) {
+        entry.children = entry.children
```
shafeeqd959, line 634: Why do we need assign this again here ? "entry.children = entry.children"

shafeeqd959:
@antonyagustine , I have looked into this code base, comments are quite useful here. I have one suggestion is that we need to be mindful when it comes to adding comments, I could see some comments which can be avoided as it is very straight forward for anyone to understand it from the code since we are using a good standard for naming and using typescripts as well. 

"The classic rule of thumb is:  Code tells you how and comments tell you why"

aman19K (approved):

antonyagustine:
> @antonyagustine , I have looked into this code base, comments are quite useful here. I have one suggestion is that we need to be mindful when it comes to adding comments, I could see some comments which can be avoided as it is very straight forward for anyone to understand it from the code since we are using a good standard for naming and using typescripts as well.
> 
> "The classic rule of thumb is: Code tells you how and comments tell you why"

Hi @shafeeqd959,

Thanks a lot for your suggestion, and I completely agree that `"The classic rule of thumb is: Code tells you how and comments tell you why"`.

However, my code might not inherently produce documentation about its functionality, even if the functions are self-explanatory. That's where my comments come into play. We should consider this from a long-term perspective. Comments aren't solely for the developer to understand the function when they look at it.

- Take, for example, someone trying to utilize my code. It should offer them a clear understanding of what the function is and what it does without needing to delve into the dist/node_module file (Like code editor/JS/TS intenseness).
- In the future, we might require or decide to integrate a documentation generator with our CI for creating developer-friendly documentation. However, until that happens, it's not feasible to go back and add comments to all functions.

@shafeeqd959, I'd recommend always adding comments to any function you write, regardless of whether the code is exceptional, self-explanatory, or seemingly straightforward. It helps in various scenarios and promotes a better understanding for everyone involved.

shafeeqd959 (approved):

PR #1230, opened 2024-01-08
Fix/cs 43366 -  Axios version bump

cs-raj:


packages/contentstack-import/package.json
```diff
@@ -9,7 +9,7 @@
     "@contentstack/cli-utilities": "~1.5.10",
     "@contentstack/management": "~1.13.0",
     "@oclif/core": "^2.9.3",
-    "axios": "^1.6.3",
+    "axios": "^1.6.4",
```
antonyagustine, line 12: Can we remove axios from dependency as it's not used in the code?

cs-raj: yes

cs-raj: done

aman19K (approved):

PR #1235, opened 2024-01-09
CS-43269-gcp changes in import

cs-raj:
version bump:
@contentstack/cli-cm-import: 1.13.0 -> 1.13.1

### DIFF
--- a/packages/contentstack-clone/package.json
+++ b/packages/contentstack-clone/package.json
@@ -6,7 +6,7 @@
   "bugs": "https://github.com/rohitmishra209/cli-cm-clone/issues",
   "dependencies": {
     "@contentstack/cli-cm-export": "~1.10.2",
-    "@contentstack/cli-cm-import": "~1.13.0",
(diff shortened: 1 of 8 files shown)


packages/contentstack-clone/package.json
```diff
@@ -6,7 +6,7 @@
   "bugs": "https://github.com/rohitmishra209/cli-cm-clone/issues",
   "dependencies": {
     "@contentstack/cli-cm-export": "~1.10.2",
-    "@contentstack/cli-cm-import": "~1.13.0",
+    "@contentstack/cli-cm-import": "~1.14.0",
```
aman19K, line 9: @cs-raj , I think in case of minor version we need to do version bump ?

aman19K (approved):

packages/contentstack-clone/package.json
cs-raj: Yes @aman19K I changed it to patch version as changes are minor

antonyagustine (approved):

PR #1249, opened 2024-01-17
CS-43112- Added GCP-NA support in CLI and removed the option from set:region command till the release happens

cs-raj:
version bump:
@contentstack/cli: 1.12.1 -> 1.13.0
@contentstack/cli-utilities: 1.5.10 -> 1.5.11
@contentstack/cli-cm-branches: 1.0.21 -> 1.0.22
@contentstack/cli-config: 1.5.1 -> 1.6.0
@contentstack/cli-cm-export: 1.10.3 -> 1.10.4
@contentstack/cli-cm-import: 1.13.1 -> 1.13.2


packages/contentstack-config/src/commands/config/set/region.ts
```diff
@@ -46,6 +46,7 @@ export default class RegionSetCommand extends BaseCommand<typeof RegionSetComman
     '$ csdx config:set:region EU',
     '$ csdx config:set:region AZURE-NA',
     '$ csdx config:set:region AZURE-EU',
+    '$ csdx config:set:region GCP-NA',
```
netrajpatel, line 49: @cs-raj This example should not be there as part of current release

packages/contentstack-config/src/utils/interactive.ts
```diff
@@ -11,6 +11,7 @@ export const askRegions = async (): Promise<string> => {
       { name: 'EU', value: 'EU' },
       { name: 'AZURE-NA', value: 'AZURE-NA' },
       { name: 'AZURE-EU', value: 'AZURE-EU' },
+      // { name: 'GCP-NA', value: 'GCP-NA' },
```
netrajpatel, line 14: @cs-raj Can you remove this line rather commenting

netrajpatel (changes requested):

packages/contentstack-config/src/commands/config/set/region.ts
cs-raj: Removed

packages/contentstack-config/src/utils/interactive.ts
cs-raj: removed this too

netrajpatel (dismissed):

shafeeqd959 (approved):

aman19K (approved):

PR #1253, opened 2024-01-22
CS-43546: Update in import command

cs-raj:
version bump: @contentstack/cli-cm-import
1.13.0 -> 1.13.1
1. Updated the global field to remove the references from RTE (Advanced and custom).
2. After the creation of content types add the create-extension so that the extensions which depend on content types are created

- Removed the extension from extensions which depend on the content type.
- Then once the content types are created then taking all the content types.
- If the title of the content type on which the extension is depending is there then adding the to the list of extensions that needed to be created.
- Creating the extension then and updating the mapper file


cs-raj:
Closing the PR as the fixes were added as a part of another ticket

PR #1289, opened 2024-02-06
CS-43741-handled why logs where not created in mentioned path

cs-raj:
version bump:
No

### DIFF
--- a/packages/contentstack-export/src/commands/cm/stacks/export.ts
+++ b/packages/contentstack-export/src/commands/cm/stacks/export.ts
@@ -98,7 +98,7 @@ export default class ExportCommand extends Command {
   static aliases: string[] = ['cm:export'];
 
   async run(): Promise<void> {
-    let exportDir: string;
+    let exportDir: string = path.join(__dirname, 'logs');
     try {
       const { flags } = await this.parse(ExportCommand);
       let exportConfig = await setupExportConfig(flags);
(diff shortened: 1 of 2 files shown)


antonyagustine (dismissed):

packages/contentstack-export/src/commands/cm/stacks/export.ts
```diff
@@ -117,8 +117,7 @@ export default class ExportCommand extends Command {
       log({ data: exportDir } as ExportConfig, `Failed to export stack content - ${formatError(error)}`, 'error');
       log(
         { data: exportDir } as ExportConfig,
-        `The log has been stored at ${
-          exportDir ? path.join(exportDir, 'logs', 'export') : path.join(__dirname, 'logs')
```
aman19K, line 121: @cs-raj , we created an export folder inside the log directory. Make sure it functions in the same manner, please.

cs-raj: Checked it it's working fine

aman19K (approved):

antonyagustine (approved):

packages/contentstack-export/src/utils/logger.ts
```diff
@@ -134,7 +134,7 @@ function init(_logPath: string) {
 }
 
 export const log = async (config: ExportConfig, message: any, type: string) => {
-  const logsPath = config.data || __dirname;
+  const logsPath = config.data;
```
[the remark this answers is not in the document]
cs-raj: the data will always be present as we are passing it every time we are creating the logs in case of error is thrown and if not then we have it for the path
`log({ data: exportDir } as ExportConfig, `The log has been stored at ${exportDir}`, 'info');`

antonyagustine (approved):

aman19K (approved):

PR #1306, opened 2024-02-22
CS-43976 - added audit for workflow

cs-raj:


packages/contentstack-audit/src/modules/workflows.ts
```diff
@@ -0,0 +1,73 @@
...
+    this.folderPath = resolve(config.basePath, config.moduleConfig[this.moduleName].dirName);
+    this.ctUidSet = new Set(['$all']);
+    this.fixWorkflowSchema = [];
+  }
+
+  run() {
+    if (!existsSync(this.folderPath)) {
+      this.log(`Skipping ${this.moduleName} audit`, 'warn');
+      this.log($t(auditMsg.NOT_VALID_PATH, { path: this.folderPath }), { color: 'yellow' });
+    }
```
antonyagustine, line 39: If the folder does not exist it should not proceed to the next line we have to break or stop the process there itself.

packages/contentstack-audit/src/modules/workflows.ts
```diff
@@ -0,0 +1,73 @@
...
+        $t(auditMsg.SCAN_WF_SUCCESS_MSG, {
+          name: workflow.name,
+          module: this.config.moduleConfig[this.moduleName].name,
+        }),
+        'info',
+      );
+    });
+    return this.fixWorkflowSchema;
+  }
+}
```
antonyagustine, line 73: How are we differentiating the audit and fix command? or is this logic only for audit not for fix?

antonyagustine (changes requested):

packages/contentstack-audit/src/modules/workflows.ts
cs-raj: It's for audit only

packages/contentstack-audit/src/audit-base-command.ts
```diff
@@ -309,6 +331,49 @@ export abstract class AuditBaseCommand extends BaseCommand<typeof AuditBaseComma
...
+          row[column] = issue[column];
+          row[column] = typeof row[column] === 'object' ? JSON.stringify(row[column]) : row[column];
+        }
+
+        if (this.currentCommand === 'cm:stacks:audit:fix') {
+          row['Fix status'] = row.fixStatus;
+        }
+
+        rowData.push(row);
+      }
```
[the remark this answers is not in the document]
cs-raj: Yes but the columns would be empty

packages/contentstack-audit/src/modules/workflows.ts
cs-raj: Got it I will add it

cs-raj: added

antonyagustine (commented):
Can you bump the version?

cs-raj:
> Can you bump the version?

Yes

antonyagustine (approved):

packages/contentstack-audit/src/modules/workflows.ts
```diff
@@ -0,0 +1,138 @@
...
+      );
+    });
+    if (this.fix && this.missingCtInWorkflows.length) {
+      await this.fixWorkflowSchema();
+    }
+    return this.missingCtInWorkflows;
+  }
+
+  async fixWorkflowSchema() {
+    for (let i in this.workflowSchema) {
```
aman19K, line 91: @cs-raj [Minor suggestion], Try to avoid i,j etc in loop.

aman19K (approved):

packages/contentstack-audit/src/modules/workflows.ts
cs-raj: Yes

antonyagustine (commented):
@cs-raj The following unit test case is failing.

`should expect fixWorkflow schema to run`

aman19K (dismissed):

antonyagustine (approved):

aman19K (approved):

PR #1315, opened 2024-02-27
CS-44158-when adding the path using the inquirer for both import and export removed the quotation marks and version bump

cs-raj:
version bump:
@contentstack/cli-cm-import: 1.14.0 -> 1.14.1
@contentstack/cli-cm-export: 1.11.0 -> 1.11.1
@contentstack/cli-utilitties: already bumped in development


antonyagustine (commented):
Have we tested the path passed in the double quote which has space?

`for instance "./temp path"`

cs-raj:
> "./temp path"

Yes, it worked

antonyagustine (approved):

aman19K (approved):

packages/contentstack-import/src/utils/import-config-handler.ts
```diff
@@ -21,10 +21,16 @@ const setupConfig = async (importCmdFlags: any): Promise<ImportConfig> => {
...
+  const pattern = /[*$%#<>{}!&?]/g;
+  while(pattern.test(config.contentDir)) {
+    console.log(`Your mentioned directory path contains special characters please add a path without them`);
+    config.contentDir = await askContentDir()
+  }
   config.contentDir = path.resolve(config.contentDir);
   //Note to support the old key
   config.data = config.contentDir;
-
+  console.log(config.contentDir)
```
antonyagustine, line 33: Do we need to keep this log?

aman19K (approved):

packages/contentstack-import/src/utils/import-config-handler.ts
cs-raj: Removed

aman19K (approved):

antonyagustine (approved):

PR #1339, opened 2024-03-11
CS-43630 - Handling the embedded entries cretion in RTE, version bump

cs-raj:
version bump:
@contentstack/cli : 1.16.0 -> 1.17.0
@contentstack/cli-cm-import: 1.14.2 -> 1.15.0
@contentstack/cli-cm-clone: 1.10.2 -> 1.10.3
@contentstack/cli-cm-seed: 1.7.2 -> 1.7.3

### DIFF
--- a/packages/contentstack-import/src/import/modules/entries.ts
+++ b/packages/contentstack-import/src/import/modules/entries.ts
@@ -56,6 +56,8 @@ export default class EntriesImport extends BaseClass {
   private autoCreatedEntries: Record<string, any>[];
   private taxonomiesPath: string;
   public taxonomies: Record<string, unknown>;
+  public rteCTs: any;
+  public rteCTsWithRef: any;
 
   constructor({ importConfig, stackAPIClient }: ModuleClassParams) {
     super({ importConfig, stackAPIClient });
@@ -85,6 +87,8 @@ export default class EntriesImport extends BaseClass {
     this.envs = {};
     this.autoCreatedEntries = [];
     this.failedEntries = [];
+    this.rteCTs = [];
+    this.rteCTsWithRef = [];
   }
 
   async start(): Promise<any> {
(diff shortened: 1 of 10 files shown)


packages/contentstack-import/src/utils/entries-helper.ts
```diff
@@ -534,11 +550,37 @@ export const restoreJsonRteEntryRefs = (
...
       }
+      case 'text': {
+        if (entry[element.uid] && element.field_metadata.rich_text_type) {
+          entry[element.uid] = sourceStackEntry[element.uid];
+          const matches: string[] = [];
+          Object.keys(uidMapper).forEach((uid) => {
+            if (sourceStackEntry[element.uid].indexOf(uid) !== -1) {
+              matches.push(uid);
+            }
+          });
```
aman19K, line 561: @cs-raj here can we use filter? For example:-

const matches = Object.keys(uidMapper).filter((uid) =>  if (sourceStackEntry[element.uid].indexOf(uid) !== -1) return uid;)

packages/contentstack-import/src/utils/entries-helper.ts
```diff
@@ -534,11 +550,37 @@ export const restoreJsonRteEntryRefs = (
...
+      }
     }
   }
   return entry;
 };
 
+function updateUids(str:string,match:string,uidMapper:Record<string, string>) {
+  return str.replace(new RegExp(match, 'g'), (match:string) => uidMapper[match])
+}
+
```
aman19K, line 583: @cs-raj formatting missing

packages/contentstack-import/src/import/modules/entries.ts
```diff
@@ -56,6 +56,8 @@ export default class EntriesImport extends BaseClass {
   private autoCreatedEntries: Record<string, any>[];
   private taxonomiesPath: string;
   public taxonomies: Record<string, unknown>;
+  public rteCTs: any;
+  public rteCTsWithRef: any;
```
aman19K, line 60: @cs-raj private will be fine here?

packages/contentstack-import/src/utils/entries-helper.ts
cs-raj: Added

packages/contentstack-import/src/utils/entries-helper.ts
cs-raj: Formatted

packages/contentstack-import/src/import/modules/entries.ts
cs-raj: While adding the test cases, It will be directly accessible

aman19K (approved):

antonyagustine (approved):

PR #1352, opened 2024-03-26
upgraded tar to 6.2.1

cs-raj:
@oclif/plugin-plugins: 4.1.9 -> 5.0.0
tar: 6.1.3 -> 6.2.1


aman19K (approved):

packages/contentstack-audit/package.json
```diff
@@ -21,7 +21,7 @@
     "@contentstack/cli-command": "~1.2.16",
     "@contentstack/cli-utilities": "~1.6.0",
     "@oclif/plugin-help": "^5",
-    "@oclif/plugin-plugins": "^4.1.9",
+    "@oclif/plugin-plugins": "^5.0.0",
```
antonyagustine, line 24: @cs-raj have you validated if there are any breaking changes with this MAJOR version? As I can see, they released so many prereleases for `v5-x.x-beta` to test, I guess.

packages/contentstack-import/package.json
```diff
@@ -34,7 +34,7 @@
     "@types/mocha": "^8.2.2",
     "@types/node": "^14.14.32",
     "@types/sinon": "^10.0.2",
-    "@types/tar": "^4.0.3",
+    "@types/tar": "^6.1.3",
```
antonyagustine, line 37: Is there any breaking changes for this package?

antonyagustine (commented):
It could be better to mention in the description about the change if we update any package to MAJOR version.

packages/contentstack-audit/package.json
cs-raj: I didn't find any breaking changes

packages/contentstack-import/package.json
cs-raj: No

antonyagustine (approved):

PR #1372, opened 2024-04-12
CS - 44008 - fixed the issue of logger path not setting up from env variable and version bump

cs-raj:
version bump:
@contentstack/cli: 1.17.2 -> 1.17.3
@contentstack/cli-cm-import: 1.15.1 -> 1.15.2


packages/contentstack-import/src/utils/log.ts
```diff
@@ -26,7 +26,8 @@ export function log(entryOrMessage: MessageType, logType?: LogsType, hidden?: bo
 
 export function initLogger(config?: ImportConfig | undefined) {
   if (!logger) {
-    const basePath = pathValidator(join(config?.data ?? process.cwd(), 'logs', 'import'));
+    const basePath =
+      process.env.CS_CLI_CONFIG_PATH || pathValidator(join(config?.data ?? process.cwd(), 'logs', 'import'));
```
aman19K, line 30: @cs-raj , Please check with the team if we support environment-based configuration paths and if so, we also need to support them in the case of the export plugin. Furthermore, please verify the clone plugin.

cs-raj: Removed the env option, we now only support config based and have the common var for both import and export the logs will include audit logs also.

packages/contentstack/package.json
```diff
@@ -10,7 +10,7 @@
   "types": "lib/index.d.ts",
   "scripts": {
     "build": "npm run clean && npm run compile",
-    "clean": "rm -rf ./lib ./node_modules tsconfig.build.tsbuildinfo",
+    "clean": "rm -rf ./li b ./node_modules tsconfig.build.tsbuildinfo",
```
aman19K, line 13: @cs-raj pls check unnecessary space will create problem

packages/contentstack-import/src/utils/logger.ts
```diff
@@ -136,14 +136,14 @@ function init(_logPath: string) {
 }
 
 export const log = async (config: ImportConfig, message: any, type: string) => {
-  config.data = config.data || path.join(__dirname, 'logs');
+  config.importExportLogPath = config.importExportLogPath || config.data || path.join(__dirname, 'logs');
```
aman19K, line 139: @cs-raj , We use config.data everywhere in the import and export plugin, and now you're using config.importExportLogPath, so hopefully you've tested.

packages/contentstack-import/src/types/import-config.ts
```diff
@@ -11,6 +11,7 @@ export interface ExternalConfig {
 }
 
 export default interface ImportConfig extends DefaultConfig, ExternalConfig {
+  importExportLogPath: string;
```
aman19K, line 14: @cs-raj it's optional param right ?

packages/contentstack-import/src/utils/logger.ts
cs-raj: Yes

packages/contentstack-import/src/types/import-config.ts
cs-raj: No, as we will be assigning it some value whether explicitly passed or the backupDir path or the present working directory

packages/contentstack/package.json
cs-raj: yes, Removed

aman19K (approved):

aman19K (approved):

shafeeqd959 (approved):

antonyagustine (approved):

PR #1375, opened 2024-04-17
fixed the issue of filtering entries using the content-type flag in both cross-publish and entries publish and version bump

cs-raj:
version bump:
@contentstack/cli: 1.17.3 -> 1.17.4
@contentstack/cli-cm-bulk-publish: 1.4.4 -> 1.4.5


cs-raj:
The changes will be released as hotfix though the PR #1376

PR #1384, opened 2024-04-22
CS - 44747 - added new script for update of references used in entries

cs-raj:
version bump:
@contentstack/cli: 1.17.3 -> 1.17.4
@contentstack/cli-migration: 1.5.3 -> 1.5.4

### DIFF
--- a/packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
+++ b/packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
@@ -1,17 +1,22 @@
 const fs = require('fs');
 const chalk = require('chalk');
+const path = require('path');
 module.exports = async ({ migration, stackSDKInstance, managementAPIClient, config }) => {
-  const modules = ['entries', 'assets', 'extensions'];
+  const modules = ['entries', 'assets', 'extensions', 'marketplace_apps'];
 
   const readAllModulesUids = (filePath) => {
-    let uidMapping = [];
+    let uidMapping = {};
+
     modules.forEach((module) => {
-      if (fs.existsSync(`${filePath}/mapper/${module}/uid-mapping.json`)) {
-        let mappedIds = fs.readFileSync(`${filePath}/mapper/${module}/uid-mapping.json`, 'utf-8');
-        mappedIds = JSON.parse(mappedIds);
-        uidMapping = { ...uidMapping, ...mappedIds };
-      } else {
-        console.log(chalk.red(`Mapper Does not exist for module '${module}'`));
+      const mappingFilePath = path.join(filePath, 'mapper', module, 'uid-mapping.json');
+      if (fs.existsSync(mappingFilePath)) {
+        const mappedIds = JSON.parse(fs.readFileSync(mappingFilePath, 'utf-8'));
+
+        if (module === 'marketplace_apps') {
+          Object.values(mappedIds).forEach((ids) => Object.assign(uidMapping, ids));
+        } else {
+          Object.assign(uidMapping, mappedIds);
+        }
       }
     });
 
@@ -22,7 +27,7 @@ module.exports = async ({ migration, stackSDKInstance, managementAPIClient, conf
     try {
       let entries = [];
       let skip = 0;
-      let limit = 1;
+      let limit = 100;
       let count = 0;
       while (skip <= count) {
         const res = await stackSDKInstance.contentType(ct).entry().query({ include_count: true, skip, limit }).find();
@@ -32,9 +37,8 @@ module.exports = async ({ migration, stackSDKInstance, managementAPIClient, conf
       }
       return entries;
     } catch (err) {
-      console.log(err);
-      console.log(chalk.red(`No entry is found for Content-type ${ct}`));
-      return [];
+      console.log(chalk.red(`Cannot Fetch Entries`));
+      throw err;
     }
   };
 
@@ -42,9 +46,27 @@ module.exports = async ({ migration, stackSDKInstance, managementAPIClient, conf
     try {
       return await stackSDKInstance.contentType(ct).fetch();
     } catch (err) {
-      console.log(chalk.red(`Content-type '${ct}' is not present in stack`));
-      return {};
+      console.log(chalk.red(`Error in Fetching the Content-type '${ct}' due to ${err.errorMessage}`));
     }
+  };
+
+  const findAllEntriesUid = (stringifiedEntry) => {
+    let pattern = /\bblt\w*/g;
+    let matches = stringifiedEntry.match(pattern);
+    return matches;
+  };
+
+  const replaceEntriesWithUpdatedUids = (matches, uidMapping, stringifiedEntry) => {
+    let isUpdated = false;
+    matches.forEach((m) => {
+      if (Object.keys(uidMapping).includes(m)) {
+        let regex = new RegExp(m, 'g');
+        stringifiedEntry = stringifiedEntry.replace(regex, uidMapping[m]);
+        console.log(chalk.green(`Replacing UID '${m}' with '${uidMapping[m]}'`));
+        isUpdated = true;
+      }
+    });
+    return { stringifiedEntry, isUpdated };
   };
 
   const updateEntryTask = () => {
@@ -63,8 +85,8 @@ module.exports = async ({ migration, stackSDKInstance, managementAPIClient, conf
           }
 
           for (let ct of config.contentTypes) {
-            let { schema: ctSchema } = await getContentTypeSchema(ct);
-            // console.log(ctSchema);
+            let { schema: ctSchema } = (await getContentTypeSchema(ct)) ?? {};
+
             if (!ctSchema) {
               continue;
             }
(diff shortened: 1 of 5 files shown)


packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,138 @@
+const fs = require('fs');
+const chalk = require('chalk');
+module.exports = async ({ migration, stackSDKInstance, managementAPIClient, config }) => {
+  const modules = ['entries', 'assets', 'extensions'];
+
```
aman19K, line 5: @cs-raj , do we need to handle market place as well ?

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,138 @@
...
+  const readAllModulesUids = (filePath) => {
+    let uidMapping = [];
+    modules.forEach((module) => {
+      if (fs.existsSync(`${filePath}/mapper/${module}/uid-mapping.json`)) {
```
aman19K, line 9: @cs-raj for path Concatenation pls use path.join or resolve instead of doing it manually

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,138 @@
+const fs = require('fs');
+const chalk = require('chalk');
+module.exports = async ({ migration, stackSDKInstance, managementAPIClient, config }) => {
+  const modules = ['entries', 'assets', 'extensions'];
+
+  const readAllModulesUids = (filePath) => {
+    let uidMapping = [];
```
aman19K, line 7: @cs-raj ,It would be preferable to use reduce rather than foreach in this case as you have declared uidMapping as an array and are expecting an object at the end.

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,138 @@
...
+    });
+
+    return uidMapping;
+  };
+
+  const getEntries = async (ct) => {
+    try {
+      let entries = [];
+      let skip = 0;
+      let limit = 1;
```
aman19K, line 25: @cs-raj , You can choose a maximum limit of 100 as an alternative to 1. This would prevent pointless API calls. For example, an API call would be made 60 times if a single CT had 60 entries. You can also provide the limit and skip value through config if required.

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,138 @@
...
+        const res = await stackSDKInstance.contentType(ct).entry().query({ include_count: true, skip, limit }).find();
+        count = res.count;
+        skip += limit;
+        entries = [...entries, ...res.items];
+      }
+      return entries;
+    } catch (err) {
+      console.log(err);
+      console.log(chalk.red(`No entry is found for Content-type ${ct}`));
+      return [];
```
aman19K, line 37: @cs-raj don't think this catch block has to be returned to. From the location where you are calling this function, you may manage this condition.

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+  };
+
+  const getContentTypeSchema = async (ct) => {
+    try {
+      return await stackSDKInstance.contentType(ct).fetch();
+    } catch (err) {
+      console.log(chalk.red(`Content-type '${ct}' is not present in stack`));
+      return {};
```
aman19K, line 46: @cs-raj don't think this catch block has to be returned to. From the location where you are calling this function, you may manage this condition.

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,138 @@
...
+
+            let entry = await getEntries(ct);
+
+            if (entry.length === 0) {
+              log(chalk.red(`No entry found for the CT Content-type '${ct}'`));
+              continue;
+            }
+            let uidMapping = readAllModulesUids(config['mapper-path']);
+
+            for (let e of entry) {
```
aman19K, line 84: @cs-raj [Minor] We can divide the logic into several functions rather than writing it all in one. so that debugging and fixing the issues would be simple and also it will increase readability.

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
cs-raj: Yes, Let's add it too

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
cs-raj: Got it Added

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
cs-raj: It's typing error, the uidMapping will be declared as an Object only

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,138 @@
...
+
+  const readAllModulesUids = (filePath) => {
+    let uidMapping = [];
+    modules.forEach((module) => {
+      if (fs.existsSync(`${filePath}/mapper/${module}/uid-mapping.json`)) {
+        let mappedIds = fs.readFileSync(`${filePath}/mapper/${module}/uid-mapping.json`, 'utf-8');
+        mappedIds = JSON.parse(mappedIds);
+        uidMapping = { ...uidMapping, ...mappedIds };
+      } else {
+        console.log(chalk.red(`Mapper Does not exist for module '${module}'`));
```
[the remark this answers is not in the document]
cs-raj: I think yes, as we will be mentioning the modules which this script will cover and in case any modules mapping is missing , we know, but still this script will run only after import if the there is no mapper for a module that means nothing got created and it's used nowhere in the entry. Let's remove it

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
cs-raj: Thanks @aman19K For pointing this out I actually missed to change it to original values

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
cs-raj: Removed it

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
cs-raj: Yes divided it into 3 different functions

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,138 @@
...
+      }
+    });
+
+    return uidMapping;
```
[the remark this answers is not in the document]
cs-raj: Yes

aman19K (dismissed):

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,156 @@
...
+  const findAllEntriesUid = (stringifiedEntry) => {
+    let pattern = /\bblt\w*/g;
+    let matches = stringifiedEntry.match(pattern);
+    return matches;
+  };
+
+  const replaceEntriesWithUpdatedUids = (matches, uidMapping, stringifiedEntry) => {
+    let isUpdated = false;
+    matches.forEach((m) => {
+      if (Object.keys(uidMapping).includes(m)) {
```
antonyagustine, line 62: `Object.keys(uidMapping)` can be out of the loop to avoid unnecessary loop as it's constant and it doesn’t require any dynamic value from the loop

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -0,0 +1,156 @@
...
+    try {
+      let entries = [];
+      let skip = 0;
+      let limit = 100;
+      let count = 0;
+      while (skip <= count) {
+        const res = await stackSDKInstance.contentType(ct).entry().query({ include_count: true, skip, limit }).find();
+        count = res.count;
+        skip += limit;
+        entries = [...entries, ...res.items];
```
antonyagustine, line 36: It seems unnecessary loop happening here `[...entries, ...res.items]`.

It can be just `entries.push(...res.items)` or `entries = entries.concat(...res.items)`

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
cs-raj: Done

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
cs-raj: Added

antonyagustine (approved):

aman19K (approved):

PR #1389, opened 2024-04-27
feat: Added Audit and audit fix for Select fields in entries and minor version bump

cs-raj:
version bump:
@contentstack/cli-audit: 1.5.4 -> 1.6.0

PR Includes:
Audit and Audit:fix for Select fields of the entries.
Since CTs can't be created without options not adding into CTs

### DIFF
--- a/packages/contentstack-audit/src/modules/entries.ts
+++ b/packages/contentstack-audit/src/modules/entries.ts
@@ -34,6 +34,8 @@ import {
   EntryReferenceFieldDataType,
   ExtensionOrAppFieldDataType,
   EntryExtensionOrAppFieldDataType,
+  EntrySelectFeildDataType,
+  SelectFeildStruct,
 } from '../types';
 import { print } from '../util';
 import GlobalField from './global-fields';
@@ -53,6 +55,7 @@ export default class Entries {
   public ctSchema: ContentTypeStruct[];
   protected entries!: Record<string, EntryStruct>;
   protected missingRefs: Record<string, any> = {};
+  protected missingSelectFeild: Record<string, any> = {};
   public entryMetaData: Record<string, any>[] = [];
   public moduleName: keyof typeof auditConfig.moduleConfig = 'entries';
   public isEntryWithoutTitleField: boolean = false;
@@ -94,6 +97,7 @@ export default class Entries {
           this.entries = entries;
 
           for (const entryUid in this.entries) {
+
             const entry = this.entries[entryUid];
             const { uid, title } = entry;
             this.currentUid = uid;
@@ -103,9 +107,13 @@ export default class Entries {
               this.missingRefs[this.currentUid] = [];
             }
 
+            if (!this.missingSelectFeild[this.currentUid]) {
+              this.missingSelectFeild[this.currentUid] = [];
+            }
             if (this.fix) {
               this.removeMissingKeysOnEntry(ctSchema.schema as ContentTypeSchemaType[], this.entries[entryUid]);
             }
+
             this.lookForReference([{ locale: code, uid, name: title }], ctSchema, this.entries[entryUid]);
             const message = $t(auditMsg.SCAN_ENTRY_SUCCESS_MSG, {
               title,
@@ -125,8 +133,7 @@ export default class Entries {
     // this.log('', 'info'); // Adding empty line
 
     this.removeEmptyVal();
-
-    return this.missingRefs;
+    return {missingEntryRefs:this.missingRefs, missingSelectFeild:this.missingSelectFeild};
   }
 
   /**
@@ -136,6 +143,11 @@ export default class Entries {
     for (let propName in this.missingRefs) {
       if (!this.missingRefs[propName].length) {
         delete this.missingRefs[propName];
+      }
+    }
+    for (let propName in this.missingSelectFeild) {
+      if (!this.missingSelectFeild[propName].length) {
+        delete this.missingSelectFeild[propName];
       }
     }
   }
@@ -224,9 +236,8 @@ export default class Entries {
       entry = this.runFixOnSchema(tree, field.schema as ContentTypeSchemaType[], entry);
     }
 
-    for (const child of field.schema ?? []) {
+    for (const child of field?.schema ?? []) {
       const { uid } = child;
-
       if (!entry?.[uid]) continue;
 
       switch (child.data_type) {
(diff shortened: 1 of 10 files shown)


packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -618,13 +651,167 @@ export default class Entries {
...
+    let arrayFeild = field;
+    console.log(arrayFeild)
+    arrayFeild.forEach((val: any) => {
+      console.log(val);
+      let t = find(selectOptions.choices, { value: val });
+      console.log(t);
+      if (!t) {
+        notPresent.push(val);
+      }
+    });
```
antonyagustine, line 810: Should we need those console logs?

Also, the for loop can be improved as follow

```typescript
const notPresent = arrayFeild.filter(
  (val) => !find(selectOptions.choices, { value: val })
);
```

packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -618,13 +651,167 @@ export default class Entries {
...
+    arrayFeild = arrayFeild.filter((val: any) => !notPresent.includes(val));
```
antonyagustine, line 811: Should we need those console logs?

Also, the for loop can be improved as follow

```typescript
const notPresent = arrayFeild.filter(
  (value) => !find(selectOptions.choices, { value })
);
```

Also last filter method is not required.

packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -279,6 +290,28 @@ export default class Entries {
...
+          if (child.hasOwnProperty('display_type')) {
+            this.missingSelectFeild[this.currentUid].push(
+              ...this.validateSelectField(
+                [...tree, { uid: field.uid, name: child.display_name, field: uid }],
+                child as SelectFeildStruct,
+                entry[uid] as any,
+              ),
+            );
+          }
+          break;
```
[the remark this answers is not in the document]
cs-raj: Removed it bought it under same case

packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -618,13 +651,167 @@ export default class Entries {
...
+
+          break;
+        case 'number':
+          if (field.hasOwnProperty('display_type')) {
+            entry[uid] = this.fixSelectField(
+              [...tree, { uid: field.uid, name: field.display_name, data_type: field.data_type }],
+              field as SelectFeildStruct,
+              entry[uid] as EntrySelectFeildDataType,
+            ) as EntrySelectFeildDataType;
+          }
```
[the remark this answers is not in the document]
cs-raj: Removed the repeated code bought it under same case

packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -618,13 +651,167 @@ export default class Entries {
...
+    if (multiple) {
+      if (min_instance && Array.isArray(field)) {
+        let obj = this.findNotPresentSelectField(field, selectOptions);
+        let  {notPresent} = obj
+        missingValues = notPresent;
+      } else if (Array.isArray(field)) {
+        let obj = this.findNotPresentSelectField(field, selectOptions);
+        let  {notPresent} = obj
+        missingValues = notPresent
+      }
```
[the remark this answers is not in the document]
cs-raj: Added

packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -618,13 +651,167 @@ export default class Entries {
...
+        let obj = this.findNotPresentSelectField(entry, selectOptions);
+        let {notPresent, filteredFeild } = obj;
+        missingValues = notPresent;
+        entry = filteredFeild;
+        if (entry.length === 0) {
+          isMissingValuePresent = true;
+          const defaultValue = selectOptions.choices.length > 0 ? selectOptions.choices[0].value : null;
+          entry.push(defaultValue);
+        }
+      }
```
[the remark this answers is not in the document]
cs-raj: Removed and added above

packages/contentstack-audit/src/modules/entries.ts
cs-raj: Modified the code, We require the last filter functions as we would need to add the entries that are present in the options of content-type

packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -619,12 +642,153 @@ export default class Entries {
...
+   * @param field It contains the value to be searched
+   * @param selectOptions It contains the options that were added in CT
+   * @returns An Array of entry containing only the values that were present in CT, An array of not present entries 
+   */
+  findNotPresentSelectField(field: any, selectOptions: any) {
+    let arrayFeild = field;
+    const notPresent = arrayFeild.filter(
+      (value: any) => !find(selectOptions.choices, { value })
+    );
+    arrayFeild = arrayFeild.filter((val: any) => !notPresent.includes(val));
```
antonyagustine, line 788: If both loops are required you can have a single for loop to accommodate it.

```typescript
let present = [];
let notPresent = [];

for (const value of arrayFeild) {
  const choice = find(selectOptions.choices, { value });
  
  if (choice) {
    present.push(choice);
  } else {
    notPresent.push(choice);
  }
}
```

antonyagustine (approved):

packages/contentstack-audit/src/modules/entries.ts
cs-raj: Modified

packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -619,12 +641,155 @@ export default class Entries {
...
+   * @returns if there is missing field returns field and path
+   * Else empty array
+   */
+  validateSelectField(tree: Record<string, unknown>[], fieldStructure: SelectFeildStruct, field: any) {
+    const { display_name, enum: selectOptions, multiple, min_instance, display_type } = fieldStructure;
+
+    let missingValues;
+
+    if (multiple) {
+      if (Array.isArray(field)) {
```
aman19K, line 675: @cs-raj do we need to mention else condition with logger?

packages/contentstack-audit/src/config/index.ts
```diff
@@ -54,6 +54,21 @@ const config = {
       'publish_details',
     ],
   },
+  OutputTableKeys : [
```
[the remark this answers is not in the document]
cs-raj: added

packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -619,12 +641,155 @@ export default class Entries {
...
+   * @param field It contains the value to be searched
+   * @param selectOptions It contains the options that were added in CT
+   * @returns An Array of entry containing only the values that were present in CT, An array of not present entries
+   */
+  findNotPresentSelectField(field: any, selectOptions: any) {
+    let present = [];
+    let notPresent = [];
+
+    for (const value of field) {
+      const choice = find(selectOptions.choices, { value });
```
aman19K, line 782: @cs-raj ,Instead of using find function to find elements in selectOptions.choices, we can create a Map from selectOptions.choices for constant-time lookup. This reduces the time complexity from O(n) to O(1) for each lookup operation.
You can use Map or Set.
  const choicesMap = new Map(selectOptions.choices.map(choice => [choice.value, choice]));

  for (const value of field) {
    const choice = choicesMap.get(value);
    if (choice) {
      present.push(choice.value);
    } else {
      notPresent.push(value);
    }
  }

packages/contentstack-audit/src/modules/entries.ts
cs-raj: No as we only have two choices

aman19K (dismissed):

antonyagustine (approved):

PR #1406, opened 2024-05-13
DX - 92 - Modified the logs for bulk publish

cs-raj:
changed remove publish_details function and changed the logs for both entries and assets publishing and version bump
version bump:
@contentstack/cli-cm-bulk-publish: 1.4.5 -> 1.4.6

### DIFF
--- a/packages/contentstack-bulk-publish/src/consumer/publish.js
+++ b/packages/contentstack-bulk-publish/src/consumer/publish.js
@@ -5,9 +5,8 @@
 const chalk = require('chalk');
 const path = require('path');
 const { formatError } = require('../util');
-const { configHandler } = require('@contentstack/cli-utilities')
-const apiVersionForNRP = '3.2'
-const nrpApiVersionWarning = `Provided apiVersion is invalid. ${apiVersionForNRP} is only supported value. Continuing with regular bulk-publish for now.`
+const apiVersionForNRP = '3.2';
+const nrpApiVersionWarning = `Provided apiVersion is invalid. ${apiVersionForNRP} is only supported value. Continuing with regular bulk-publish for now.`;
 
 const { getLoggerInstance, addLogs, getLogsDirPath } = require('../util/logger');
 const logsDir = getLogsDirPath();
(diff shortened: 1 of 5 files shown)


packages/contentstack-bulk-publish/src/consumer/publish.js
```diff
@@ -25,13 +24,29 @@ function initializeLogger(fileName) {
 /* eslint-disable camelcase */
 function removePublishDetails(elements) {
   if (elements && elements.length > 0) {
-    return elements.map(({ _publish_details, ...rest }) => rest);
+    elements.map((element) => {
+      if (element.hasOwnProperty('publish_details')) {
+        delete element.publish_details;
+      }
+      return element;
+    });
```
aman19K, line 32: @cs-raj , [Minor]deleted line also work you just need to use publish_details instead of _publish_details.

packages/contentstack-bulk-publish/src/consumer/publish.js
```diff
@@ -25,13 +24,29 @@ function initializeLogger(fileName) {
...
+  sanitizedData.forEach((entry) => {
+    console.log(chalk.green(`Entry UID '${entry.uid}' of CT '${entry.content_type}' in locale '${entry.locale}'`));
+  });
+}
+
+function displayAssetsDetails(sanitizedData) {
+  sanitizedData.forEach((asset) => {
+    console.log(chalk.green(`Asset UID '${asset.uid}' ${asset.version ? `and version '${asset.version}'` : ''} ${asset.locale ? `in locale '${asset.locale}'` : ''}`));
+  });
+}
```
aman19K, line 49: @cs-raj , formatting missing

aman19K (changes requested):

packages/contentstack-bulk-publish/src/consumer/publish.js
cs-raj: Used

packages/contentstack-bulk-publish/src/consumer/publish.js
cs-raj: Added, not sure how only this line it was left

PR #1409, opened 2024-05-14
fix: issue with import entries arising from compact library

aman19K:
- [x] issue with import entries arising from compact library
- [x] version bump already done

### DIFF
--- a/packages/contentstack-import/src/utils/asset-helper.ts
+++ b/packages/contentstack-import/src/utils/asset-helper.ts
@@ -375,7 +375,7 @@ function findFileUrls(schema: any, _entry: any, assetUrls: any) {
 }
 
 function updateFileFields(
-  objekt: any,
+  object: any,
   parent: any,
   pos: any,
   mappedAssetUids: any,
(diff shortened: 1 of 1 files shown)


cs-raj (dismissed):

packages/contentstack-import/src/utils/asset-helper.ts
```diff
@@ -434,6 +434,7 @@ function updateFileFields(
     for (let i = 0; i <= objekt.length; i++)
       updateFileFields(objekt[i], objekt, i, mappedAssetUids, matchedUids, unmatchedUids);
 
-    parent[pos] = _.compact(objekt);
+    // No need for _.compact() since you want to keep zero values
+    parent[pos] = _.filter(objekt, (value: any) => value !== undefined && value !== null);
```
shafeeqd959, line 438: @aman19K since we are touching this code can we rename this variable from "objekt" to an appropriate name.

aman19K: done

cs-raj (approved):

shafeeqd959 (approved):

PR #1422, opened 2024-06-03
fix: Handled pagination during the script creation

cs-raj:
version bump:
@contentstack/cli: 1.18.2 -> 1.19.0
@contentstack/cli-cm-branches: 1.0.26 -> 1.1.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24876,13 +24876,13 @@
     },
     "packages/contentstack": {
       "name": "@contentstack/cli",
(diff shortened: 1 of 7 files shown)


shafeeqd959 (commented):
@cs-raj this could be minor update, we are not fixing anything rather we are adding a feature so it should be minor update

aman19K (approved):

aman19K (approved):

PR #1423, opened 2024-06-04
DX - 647 - fixed the bug with rich text type and bugs in entries helper

cs-raj:


packages/contentstack-import/src/utils/entries-helper.ts
```diff
@@ -62,8 +62,8 @@ export const lookupEntries = function (
       if (_entry && _parent[j]) {
         if (j === len - 1 && _entry[_parent[j]]) {
           if (form_id !== '_assets') {
-            if (_entry[_parent[j]]?.length) {
-              _entry[_parent[j]].forEach((item: any, idx: any) => {
+            if (_entry[_parent[j]]?.length && Object.keys(_entry).includes(_parent[j])) {
+              _entry[_parent[j]]?.forEach((item: any, idx: any) => {
```
aman19K, line 66: @cs-raj , The Object.keys(_entry).includes(_parent[j]) check is redundant since accessing _entry[_parent[j]] will return undefined if the key does not exist, which is already handled by ?.length.

cs-raj: There was an issue with _entry[parent[j]] where the value of the parent[j] was 'filter' due to which it was not returning instead it was giving filter function. Due to which it was failing on for each

aman19K: @cs-raj , with optional chaining also?

aman19K (approved):

PR #1425, opened 2024-06-05
Implemented compass app integration with Seed and Bootstrap command

sunil-lakshman:


packages/contentstack-bootstrap/src/bootstrap/utils.ts
```diff
@@ -189,6 +189,24 @@ const envFileHandler = async (
       }\nCONTENTSTACK_LIVE_PREVIEW=${livePreviewEnabled}\nCONTENTSTACK_LIVE_EDIT_TAGS=false`;
       result = await writeEnvFile(content, filePath);
       break;
+    case 'compass-app':
+      fileName = '.env';
+      filePath = pathValidator(path.join(clonedDirectory, fileName));
```
aman19K, line 194: @sunil-lakshman , could you please verify with @cs-raj  if it is preferable to use his utility function rather than pathValidator since he has developed one to address SRE issues?
If possible try to update everywhere in **bootstrap/utils.ts**

aman19K (changes requested):

PR #1455, opened 2024-07-03
DX - 77: Added the missing modules handle function, upgraded uuid and version bump

cs-raj:
version bump:
@contentstack/cli: 1.20.0 -> 1.21.0
@contentstack/cli-migration: 1.5.6 -> 1.6.0

### DIFF
--- a/packages/contentstack-migration/src/commands/cm/stacks/migration.js
+++ b/packages/contentstack-migration/src/commands/cm/stacks/migration.js
@@ -25,7 +25,7 @@ const {
 const { ApiError, SchemaValidator, MigrationError, FieldValidator } = require('../../../validators');
 
 // Utils
-const { map: _map, constants, safePromise, errorHelper } = require('../../../utils');
+const { map: _map, constants, safePromise, errorHelper, installModules } = require('../../../utils');
 // Properties
 const { get, set, getMapInstance, resetMapInstance } = _map;
 const {
@@ -67,7 +67,7 @@ class MigrationCommand extends Command {
       this.exit();
     }
 
-    if (!filePath) {
+    if (!filePath || !fs.existsSync(filePath)) {
       this.log('Please provide the migration script file path, use --file-path flag');
       this.exit();
     }
@@ -128,6 +128,10 @@ class MigrationCommand extends Command {
 
     set(MANAGEMENT_SDK, mapInstance, stackSDKInstance);
     set(MANAGEMENT_CLIENT, mapInstance, APIClient);
+
+    if (!(await installModules(filePath, multi))) {
+      this.log(`Cannot proceed with migration as there will be module not found issue as package.json is not created`);
+    }
 
     if (multi) {
       await this.execMultiFiles(filePath, mapInstance);
--- a/packages/contentstack-migration/src/commands/cm/stacks/migration.js
+++ b/packages/contentstack-migration/src/commands/cm/stacks/migration.js
@@ -131,6 +131,7 @@ class MigrationCommand extends Command {
 
     if (!(await installModules(filePath, multi))) {
       this.log(`Failed to install the dependencies of the given scripts.`);
+      process.exit(1);
     }
 
     if (multi) {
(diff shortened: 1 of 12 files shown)


packages/contentstack-migration/src/utils/modules.js
```diff
@@ -0,0 +1,172 @@
...
+  });
+
+  return [...dependencies];
+}
+
+function createPackageJson(directory) {
+  const templateString = `{
+    "name": "MigrationPackage",
+    "version": "1.0.0",
+    "main": "1.js",
```
aman19K, line 84: @cs-raj , any specific reason for mentioning 1.js in main.

aman19K (changes requested):

packages/contentstack-migration/src/utils/modules.js
cs-raj: No, It was just the file with which it was added, shall I remove it?

cs-raj: Removed

packages/contentstack-migration/src/utils/modules.js
```diff
@@ -0,0 +1,172 @@
...
+  'trace_events',
+  'tty',
+  'url',
+  'util',
+  'v8',
+  'vm',
+  'wasi',
+  'worker_threads',
+  'zlib',
+]);
```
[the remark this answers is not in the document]
cs-raj: Thanks, Used the second one

aman19K (dismissed):

packages/contentstack-migration/src/utils/modules.js
```diff
@@ -0,0 +1,134 @@
...
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
+    "description": ""
+  }`;
+
+  fs.writeFileSync(path.join(sanitizePath(directory), 'package.json'), templateString);
+}
+
+function installDependencies(dependencies, directory) {
```
shafeeqd959, line 57: How do we get the version of a dependency ? Installing the latest version of a dependency would be an issue here ??

packages/contentstack-migration/src/utils/modules.js
```diff
@@ -0,0 +1,134 @@
...
+      createPackageJson(dirPath);
+    }
+
+    installDependencies(dependencies, dirPath);
+  } else {
+    console.log(`You don't have write permission to the directory`);
+    return false;
+  }
+
+  console.log(`All modules installed successfully`);
```
shafeeqd959, line 107: "All dependencies installed successfully"

packages/contentstack-migration/src/commands/cm/stacks/migration.js
```diff
@@ -129,6 +129,10 @@ class MigrationCommand extends Command {
     set(MANAGEMENT_SDK, mapInstance, stackSDKInstance);
     set(MANAGEMENT_CLIENT, mapInstance, APIClient);
 
+    if (!(await installModules(filePath, multi))) {
+      this.log(`Cannot proceed with migration as there will be module not found issue as package.json is not created`);
```
shafeeqd959, line 133: "Failed to install the dependencies of the given scripts."  please change the text , then we can exit

shafeeqd959 (dismissed):

packages/contentstack-migration/src/utils/modules.js
cs-raj: We don't get the versions of the dependency we try to install the latest version for our use case of entry migration that won't be a problem untill we release major version. Asking or letting the user specify specific version will be as good as creating package.json manually or running npm i with specific version. The issue might arise with user running their own script with specific version

packages/contentstack-migration/src/utils/modules.js
cs-raj: Got it I will add it

packages/contentstack-migration/src/commands/cm/stacks/migration.js
cs-raj: added

aman19K (approved):

PR #1469, opened 2024-07-22
Compass app integration and removed error messages while importing a stack

sunil-lakshman:
…stack.


packages/contentstack-bootstrap/src/bootstrap/utils.ts
```diff
@@ -34,6 +34,40 @@ export const setupEnvironments = async (
...
+          ],
+          "expires_on": "3000-01-01",
+          "is_email_notification_enabled":false
+      }
+  }
+  managementTokenResult = await managementAPIClient
+    .stack({ api_key: api_key })
+    .managementToken()
+    .create(managementBody);
+  }
```
aman19K, line 70: @sunil-lakshman , Is the creation of a management token required? We can throw errors in place of creating management tokens. We're attempting to create management tokens without user permission, and I believe this is going to be a major problem.
cc: @netrajpatel @shafeeqd959

shafeeqd959: @aman19K for other apps we are creating delivery token internally, Bootstrap flow should setup all the required environment values internally.

@sunil-lakshman Correct if I am wrong, My understanding here is that we need to have management token here to make this app working.

sunil-lakshman: @shafeeqd959 Management token required for compass app

packages/contentstack-bootstrap/src/config.ts
```diff
@@ -50,7 +51,8 @@ const config: Configuration = {
     },
     'compass-app': {
       source: 'SunilLsagar/universal-demo',
```
shafeeqd959, line 53: Why is this code and stacks is in your repo ?

packages/contentstack-seed/src/seed/index.ts
```diff
@@ -180,6 +182,11 @@ export default class ContentModelSeeder {
 
   async shouldProceed(api_key: string) {
```
shafeeqd959, line 183: Bump seed command as well, minor version bump

packages/contentstack/package.json
```diff
@@ -24,13 +24,13 @@
   "dependencies": {
     "@contentstack/cli-audit": "~1.6.3",
     "@contentstack/cli-auth": "~1.3.19",
-    "@contentstack/cli-cm-bootstrap": "~1.9.3",
+    "@contentstack/cli-cm-bootstrap": "~1.10.0",
     "@contentstack/cli-cm-branches": "~1.1.1",
     "@contentstack/cli-cm-bulk-publish": "~1.4.7",
     "@contentstack/cli-cm-export": "~1.11.5",
```
shafeeqd959, line 30: Minor version bump required in seed command, also in documentation we need to add the --locale flag in seed command doc

packages/contentstack-bootstrap/src/config.ts
sunil-lakshman: @shafeeqd959 Still compass app source code repo and stacks repo's are private, Once it's public will replace it.

packages/contentstack-bootstrap/src/bootstrap/utils.ts
aman19K: @shafeeqd959 , creating management token without user consent. isn't it going to be problem? May be some prompt for confirmation so that they'll aware about it.

shafeeqd959: @aman19K,  what about other apps where we create delivery tokens ?

shafeeqd959: This point is valid,  we have to make a decision that, is it just required here or in other apps as well, if needed we can have one ticket to make that change.

aman19K (approved):

PR #1474, opened 2024-07-23
fixed the SRE issues

cs-raj:
version bump:
Will be done once all the changes are verified and stage doesn't break


packages/contentstack-import/src/utils/asset-helper.ts
```diff
@@ -247,31 +247,56 @@ export const lookupAssets = function (
   assetUids = _.uniq(assetUids);
   assetUrls = _.uniq(assetUrls);
   let entry = JSON.stringify(data.entry);
+  
+  // Note: Use this code in case production breaks
```
shafeeqd959, line 251: @cs-raj if the production code breaks, how client is supposed to enable this code ?

shafeeqd959 (commented):
Only tested proven can be accepted here,  for bypassing SRE we cant break the logic, This is not acceptable. Also why we need to keep unnecessary code as commented

"Note: Use this incase the production breaks"

aman19K (commented):
@cs-raj , audit changes looks good but we can't take risk with asset & entries as we don't all cases. Create a ticket and share with SRE team that we can't fix these issues.

packages/contentstack-import/src/utils/entries-helper.ts
```diff
@@ -201,20 +203,51 @@ export const lookupEntries = function (
...
+        if ((index === 0 || !isWordCharacter(entry.charAt(index - 1))) &&
+            (index + sanitizedUid.length === entry.length || !isWordCharacter(entry.charAt(index + sanitizedUid.length)))) {
+          entry = entry.substring(0, index) + escapedMappedUid + entry.substring(index + sanitizedUid.length);
+          index += escapedMappedUid.length;  
+        } else {
+          index += sanitizedUid.length;  
+        }
+        index = entry.indexOf(sanitizedUid, index); 
       }
+      mapped.push(uid);
```
aman19K, line 238: @cs-raj , I think this code is more complex and harder to read compared to the regex approach. Might be it'll create problem or less efficient in case of larger strings.
This word char can be error prone.

packages/contentstack-import/src/utils/asset-helper.ts
```diff
@@ -247,31 +247,56 @@ export const lookupAssets = function (
...
+  //     //NOTE - This code was added to resolve the SRE issue but once the code was merged Assets URLs in JSON RTE started breaking
+  //     // const sanitizedUrl = escapeRegExp(assetUrl).replace(/\.\./g, '\\$&');
+  //     // const escapedMappedUrl = escapeRegExp(mappedAssetUrl).replace(/\.\./g, '\\$&');
+  //     // entry = entry.replace(new RegExp(sanitizedUrl, 'img'), escapedMappedUrl);
+  //     entry = entry.replace(new RegExp(assetUrl, 'img'), mappedAssetUrl);
+  //     matchedUrls.push(mappedAssetUrl);
+  //   } else {
+  //     unmatchedUrls.push(assetUrl);
+  //   }
+  // });
```
aman19K, line 264: @cs-raj , remove commented code

packages/contentstack-import/src/utils/asset-helper.ts
```diff
@@ -247,31 +247,56 @@ export const lookupAssets = function (
...
+  //     if (status === 'safe') {
+  //       entry = entry.replace(regex, uid);
+  //       matchedUids.push(assetUid);
+  //     }
+  //   } else {
+  //     unmatchedUids.push(assetUid);
+  //   }
+  // });
+
+  // To be deleted when prod breaks
```
aman19K, line 293: @cs-raj , remove commented code

aman19K (changes requested):

packages/contentstack-import/src/utils/entries-helper.ts
shafeeqd959: @cs-raj Why splitting logic used above is not possible here ?
@cs-raj Why we have to remove the regex implementation, what exactly error here ?

packages/contentstack-import/src/utils/asset-helper.ts
cs-raj: Yes

packages/contentstack-import/src/utils/asset-helper.ts
cs-raj: Removed

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
```diff
@@ -73,10 +75,10 @@ module.exports = async ({ migration, stackSDKInstance, managementAPIClient, conf
     let oldUids = Object.keys(uidMapping);
     matches.forEach((m) => {
       if (oldUids.includes(m)) {
-        let regex = new RegExp(m, 'g');
+        let sanitizedUid = m; 
         let { status } = validateRegex(regex);
         if (status === 'safe') {
```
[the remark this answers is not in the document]
cs-raj: Correct

packages/contentstack-import/src/utils/entries-helper.ts
```diff
@@ -580,9 +613,9 @@ export const restoreJsonRteEntryRefs = (
 
 function updateUids(str: string, match: string, uidMapper: Record<string, string>) {
   const sanitizedMatch = escapeRegExp(match);
-  const regex = new RegExp(`\\b${sanitizedMatch}\\b`, 'g');
-  let { status } = validateRegex(regex);
-  if (status === 'safe') return str.replace(regex, (matchedString) => uidMapper[matchedString]);
+  const parts = str.split(sanitizedMatch);
+  let result = parts.join(uidMapper[match] || sanitizedMatch);
```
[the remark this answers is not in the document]
cs-raj: Done

cs-raj:
> Have we tested the code ?

Yes @shafeeqd959

packages/contentstack-import/src/utils/entries-helper.ts
cs-raj: Removed and replaced with older code

packages/contentstack-migration/examples/05-Update-reference-entry-from-mapper.js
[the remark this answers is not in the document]
cs-raj: Removed it

packages/contentstack-migration/src/utils/modules.js
```diff
@@ -70,8 +70,11 @@ function installDependencies(dependencies, directory) {
 
 function executeShellCommand(command, directory = '') {
   try {
-    execSync(command, { stdio: 'inherit', cwd: directory });
-    console.log(`Command executed successfully: ${command}`);
+    if (command.startsWith('npm i')) {
+      const [cmd, ...args] = command.split(' ');
+      execFileSync(cmd, args, { stdio: 'inherit', cwd: directory });
+      console.log(`Command executed successfully: ${command}`);
+    }
```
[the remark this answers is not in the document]
cs-raj: Using execSync then issue is raised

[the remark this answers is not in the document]
aman19K: @cs-raj , for SRE fix we can't ignore and break existing code. correct ?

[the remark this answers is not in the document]
cs-raj: Correct

shafeeqd959 (approved):

packages/contentstack-import/src/utils/entries-helper.ts
```diff
@@ -201,20 +203,51 @@ export const lookupEntries = function (
...
+        }
+        index = entry.indexOf(sanitizedUid, index); 
       }
+      mapped.push(uid);
     } else {
       unmapped.push(uid);
     }
   });
+  // Note: Delete when prod breaks
+  function escapeRegExp(str:string) {
```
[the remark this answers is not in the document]
cs-raj: None the code changes are removed

aman19K (approved):

PR #1515, opened 2024-08-13
Implemented variants entries bulk publish feature

sunil-lakshman:


packages/contentstack-bulk-publish/src/commands/cm/entries/publish.js
```diff
@@ -244,6 +246,11 @@ PublishEntriesCommand.flags = {
   }),
   'delivery-token': flags.string({ description: 'Delivery token for source environment' }),
   'source-env': flags.string({ description: 'Source environment' }),
+  'include-variants': flags.boolean({ 
+    default: false, // set the default value to false
+    allowNo: true, // this allows the flag to be negated with `--no-include-variants`
```
aman19K, line 251: @sunil-lakshman , this isn't required as we're using default value.

packages/contentstack-bulk-publish/src/producer/publish-entries.js
```diff
@@ -37,14 +42,31 @@ async function getEntries(stack, contentType, locale, bulkPublish, environments,
...
+        for (let index = 0; index < entries.length; index++) {
+          let variants = [];
+
+          if (variantsFlag) {
+            const variantsEntriesResponse = await stack
+              .contentType(contentType)
+              .entry(entries[index].uid)
+              .variants()
+              .query(queryParams)
+              .find();
```
aman19K, line 55: @sunil-lakshman , then & catch block is missing.  Please create a separate function and manage pagination as well, rather than adding variation code to the same block.
if(variantsFlag){
   variants = await getVariantEntries();
}
async function getVariantEntries(){
// handle then & catch
}

packages/contentstack-bulk-publish/src/producer/publish-entries.js
```diff
@@ -37,14 +42,31 @@ async function getEntries(stack, contentType, locale, bulkPublish, environments,
...
+          }
+
           if (bulkPublish) {
             if (bulkPublishSet.length < 10) {
               bulkPublishSet.push({
                 uid: entries[index].uid,
                 content_type: contentType,
                 locale,
                 publish_details: entries[index].publish_details || [],
+                variants: variants
```
aman19K, line 69: @sunil-lakshman , [Minor], we can directly use **variants** as param instead of **variants:variants**

packages/contentstack-bulk-publish/src/producer/publish-entries.js
```diff
@@ -149,6 +171,11 @@ async function start(
     }
     process.exit(0);
   });
+
+  if (includeVariantsFlag) {
+    apiVersion = VARIANTS_PUBLISH_API_VERSION;
+  }
```
aman19K, line 177: @sunil-lakshman , do we need apiVersion here? As we've already assigned apiVersion before calling variant entries.
https://github.com/contentstack/cli/pull/1515/files#diff-4f8d1fcd74a4112f2ea9ee28af38f2f1e531bd62367a2076ae6ce6962c8c7159R33

aman19K (changes requested):

packages/contentstack-bulk-publish/src/producer/publish-entries.js
sunil-lakshman: @aman19K apiVersion is optional parameter passed through params in SDK query.

packages/contentstack-bulk-publish/src/producer/publish-entries.js
sunil-lakshman: @aman19K The payload is constructed explicitly with the required values to ensure clarity when sending it for publishing variant entries. This approach maintains consistency in the payload structure, especially with multiple parameters.

packages/contentstack-bulk-publish/src/producer/publish-entries.js
```diff
@@ -92,13 +113,32 @@ async function getEntries(stack, contentType, locale, bulkPublish, environments,
...
+      .query(queryParams)
+      .find();
+
+    const variants = variantsEntriesResponse.items.map(entry => ({
+      uid: entry.variants_uid,
+    }));
+    
+    return variants;
+  } catch (error) {
+    throw new Error(`Error fetching variants: ${error.message}`);
```
aman19K, line 138: @sunil-lakshman , From api side we get error message in errorMessage, message and errors properties so it's better to handle like this
error?.errorMessage || error?.message || error. Pls check other plugin as well

packages/contentstack-bulk-publish/src/producer/publish-entries.js
```diff
@@ -29,6 +30,10 @@ async function getEntries(stack, contentType, locale, bulkPublish, environments,
       include_publish_details: true,
     };
 
+    if (variantsFlag) {
+      queryParams.apiVersion = VARIANTS_PUBLISH_API_VERSION;
```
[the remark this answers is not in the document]
shafeeqd959: ?

shafeeqd959 (approved):

packages/contentstack-bulk-publish/src/producer/publish-entries.js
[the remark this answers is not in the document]
sunil-lakshman: yes @shafeeqd959, as per the documentation, the variants publish API requires API version 3.2.

PR #1521, opened 2024-08-19
Implemented variants entries bulk unpublish

sunil-lakshman:


packages/contentstack-bulk-publish/src/commands/cm/entries/unpublish.js
```diff
@@ -19,6 +19,7 @@ class UnpublishCommand extends Command {
     unpublishFlags.onlyAssets = false;
     unpublishFlags.onlyEntries = true;
     unpublishFlags.apiVersion = unpublishFlags['api-version'] || '3';
+    unpublishFlags.includeVariants = unpublishFlags['include-variants'] || unpublishFlags.includeVariants || false;
```
shafeeqd959, line 22: unpublishFlags['include-variants'] || unpublishFlags.includeVariants || false;

Do we really need to all different cases here "unpublishFlags['include-variants'] " this will be enough right, we are setting the default value as "false" in the below flag configuration itself, we can avoid second and third or case here

packages/contentstack-seed/package.json
```diff
@@ -22,7 +22,7 @@
     "@types/node": "^14.14.32",
     "@types/tar": "^6.1.3",
     "@types/tmp": "^0.2.0",
-    "axios": "^1.6.4",
+    "axios": "^1.7.4",
```
shafeeqd959, line 25: why did we bump the seed here, not the bulk publish, version bump can be done once we create PR  to development branch

sunil-lakshman: @shafeeqd959   SCA scan workflow is failing due to axios package bump so, i have upgraded the axios version

packages/contentstack-bulk-publish/src/producer/unpublish.js
```diff
@@ -230,6 +242,54 @@ async function getSyncEntries(
...
+}
+
+async function getVariantEntries(stack, contentType, entries, queryParams, skip = 0) {
+  try {
+    let variantQueryParams = {
+      locale: queryParams.locale || 'en-us',
+      include_count: true,
+      skip: skip,  // Adding skip parameter for pagination
+      limit: 100,  // Set a limit to fetch up to 100 entries per request
+      include_publish_details: true,
```
aman19K, line 261: @sunil-lakshman , [Minor] include_publish_details isn't required pls remove it.

aman19K (approved):

aman19K (approved):

PR #1544, opened 2024-09-04
dx | 1112 rate limit set command

harshithad0703:
- added set rate limit command
- version bump to be made
- 1 case to be checked properly (setting utilize to given limit name as comma separated string in flag)


packages/contentstack-config/src/commands/config/set/rate-limit.ts
```diff
@@ -0,0 +1,80 @@
...
+    const { flags } = await this.parse(RateLimitSetCommand);
+    const config: RateLimitConfig = {};
+
+    let { org, utilize, 'limit-name': limitName, default: isDefault } = flags;
+    if (!org) {
+      org = await askOrgID();
+    }
+    config.org = org;
+    if (isDefault) {
+      config.utilize = 50;
```
aman19K, line 60: @harshithad0703 , As we're using default value in flag then we don't need to set it here

packages/contentstack-config/src/commands/config/set/rate-limit.ts
```diff
@@ -0,0 +1,80 @@
...
+      config['limit-name'] = ['getLimit', 'limit', 'bulkLimit'];
```
aman19K, line 61: @harshithad0703 , Keep the limit-name flag value in an alternative file for usage in the flag option and default flag scenarios.

packages/contentstack-config/src/commands/config/set/rate-limit.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    const limitHandler = new RateLimitHandler();
+    const managementAPIClient = await managementSDKClient(config);
+    limitHandler.setClient(managementAPIClient);
+
+    try {
+      await limitHandler.setRateLimit(config);
+    } catch (error) {
+      cliux.error(`error : Couldn't find the organization with UID: ${org}`);
```
aman19K, line 77: @harshithad0703 , Error message should be like this
 cliux.error(`Error : Something went wrong while setting rate limit for org: ${org}`, error);

packages/contentstack-config/src/utils/rate-limit-handler.ts
```diff
@@ -0,0 +1,55 @@
+import { cliux, configHandler } from '@contentstack/cli-utilities';
+
+const requiredLimitsArray = ['getLimit', 'limit', 'bulkLimit'];
```
aman19K, line 3: @harshithad0703 ,
Keep this in an alternative file for usage in the flag option and default flag scenarios.

packages/contentstack-config/src/utils/rate-limit-handler.ts
```diff
@@ -0,0 +1,55 @@
...
+      rateLimit[config.org] = limitsToUpdate;
+      configHandler.set('rateLimit', rateLimit);
+      cliux.success('Rate limit has been set successfully');
+    } catch (error) {
+      cliux.error(`Error: Couldn't set the rate limit. ${error.message}`);
+    }
+  }
+}
+
+export { client };
```
aman19K, line 55: @harshithad0703 , why are you exporting client ?

packages/contentstack-config/src/utils/rate-limit-handler.ts
```diff
@@ -0,0 +1,55 @@
...
+            ...limitsToUpdate[limitName],
+            utilize: config.utilize,
+          };
+        }
+      }
+      rateLimit[config.org] = limitsToUpdate;
+      configHandler.set('rateLimit', rateLimit);
+      cliux.success('Rate limit has been set successfully');
+    } catch (error) {
+      cliux.error(`Error: Couldn't set the rate limit. ${error.message}`);
```
aman19K, line 50: @harshithad0703 , Are you sure that every time we get message in error? It's better to log like that
error?.errorMessage || error?.message || error

packages/contentstack-config/src/utils/rate-limit-handler.ts
```diff
@@ -0,0 +1,55 @@
...
+        if (requiredLimitsArray.includes(limitName)) {
+          limitsToUpdate[limitName] = {
+            ...limitsToUpdate[limitName],
+            utilize: config.utilize,
+          };
+        }
+      }
+      rateLimit[config.org] = limitsToUpdate;
+      configHandler.set('rateLimit', rateLimit);
+      cliux.success('Rate limit has been set successfully');
```
aman19K, line 48: @harshithad0703 , mention org in message

packages/contentstack-config/src/utils/rate-limit-handler.ts
```diff
@@ -0,0 +1,55 @@
...
+      }
+
+      for (const limitName of limitNames) {
+        if (requiredLimitsArray.includes(limitName)) {
+          limitsToUpdate[limitName] = {
+            ...limitsToUpdate[limitName],
+            utilize: config.utilize,
+          };
+        }
+      }
```
aman19K, line 45: @harshithad0703 , as you've already prepared config value in line no 28 then why are you again doing it? 
Try to handle default and custom case in single iteration

packages/contentstack-config/src/commands/config/set/rate-limit.ts
```diff
@@ -0,0 +1,80 @@
...
+import { BaseCommand } from '../../../base-command';
+import { askOrgID } from '../../../utils/interactive';
+
+interface RateLimitConfig {
+  org?: string;
+  utilize?: number;
+  limitName?: string[];
+  default?: boolean;
+  auth_token?: string;
+}
```
aman19K, line 12: @harshithad0703 ,
Since we have an interface folder, you can move the type definition there instead of keeping it within the command.
do we really need auth_token?

packages/contentstack-config/src/commands/config/set/rate-limit.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    utilize: flags.string({
+      description: 'Provide the utilization percentage',
+      default: '50',
+    }),
+
+    'limit-name': flags.string({
+      description: '[Optional] Provide the limit names separated by comma',
+      multiple: true,
+    }),
```
aman19K, line 30: @harshithad0703 ,
Provide option here similar like. 
 https://github.com/contentstack/cli/pull/1544/files#diff-59a0e4886e4b8e0671ef051c352241f147b2c3324b7a2a6cef3e9d0a68eb4d9dR27-R30
 
In options you can provide ['limit', 'getLimit', 'bulkLimit']. Keep it in some file so that we can use it everywhere. Otherwise user can provide any name

packages/contentstack-config/src/utils/rate-limit-handler.ts
```diff
@@ -0,0 +1,55 @@
+import { cliux, configHandler } from '@contentstack/cli-utilities';
+
+const requiredLimitsArray = ['getLimit', 'limit', 'bulkLimit'];
+const rateLimit = configHandler.get('rateLimit');
+
+let client: any = {};
```
aman19K, line 6: @harshithad0703 , don't think it's required

aman19K (changes requested):

aman19K (approved):

shafeeqd959 (approved):

PR #1569, opened 2024-09-18
Fixed cross publish issue and updated variant entry publish payload

sunil-lakshman:
1.  Fixed cross publish issue
2. updated variant entry publish payload
3. Upgrade the contentstack/management package v1.18.0


packages/contentstack-bulk-publish/src/consumer/publish.js
```diff
@@ -247,7 +247,7 @@ async function performBulkPublish(data, _config, queue) {
       payload['details'] = conf;
       if (bulkPublishObj.apiVersion) {
         if (!isNaN(bulkPublishObj.apiVersion) && bulkPublishObj.apiVersion === apiVersionForNRP) {
-          payload['api_version'] = bulkPublishObj.apiVersion;
+          stack.stackHeaders['api_version'] = bulkPublishObj.apiVersion;
```
aman19K, line 250: @sunil-lakshman , previously it's working with payload for NRP. can you pls check whether changes required or not? And also don't think it's good idea to send payload in this format. If it's required then it should be handled from sdk side.
CC: @netrajpatel @shafeeqd959

packages/contentstack-bulk-publish/src/producer/cross-publish.js
```diff
@@ -210,7 +210,7 @@ async function getSyncEntries(
       if (variantsFlag) {
         for (let index = 0; index < entriesResponse?.items?.length; index++) {
           let variants = [];
-          variants = await getVariantEntries(stack, entries[index].content_type_uid, entries, index, queryParamsObj);
+          variants = await getVariantEntries(stack, entriesResponse.items[index].content_type_uid, entriesResponse, index, queryParamsObj);
```
aman19K, line 213: @sunil-lakshman , [Minor] instead of writing redundant entriesResponse.items[index]. I think its better to store like this
const entries = entriesResponse.items[index];
  variants = await getVariantEntries(stack, entries.content_type_uid, entriesResponse, index, queryParamsObj);

packages/contentstack-bulk-publish/src/producer/cross-publish.js
```diff
@@ -271,7 +271,7 @@ async function getVariantEntries(stack, contentType, entries, index, queryParams
 
     const variantsEntriesResponse = await stack
       .contentType(contentType)
-      .entry(entries[index].data.uid)
+      .entry(entries.items[index].data.uid)
```
aman19K, line 274: @sunil-lakshman ,[minor] same apply here

packages/contentstack-bulk-publish/src/commands/cm/entries/publish.js
```diff
@@ -22,7 +22,7 @@ class PublishEntriesCommand extends Command {
       entriesFlags['publish-all-content-types'] || entriesFlags.publishAllContentTypes || false;
     entriesFlags.apiVersion = entriesFlags['api-version'] || '3';
     entriesFlags.includeVariants = entriesFlags['include-variants'] || entriesFlags.includeVariants || false;
-    entriesFlags.entryUid = entriesFlags['entry-uid'] || entriesFlags.entryUid;
+    entriesFlags.entryUid = entriesFlags['entry-uid'] || entriesFlags.entryUid || false;
```
aman19K, line 25: @sunil-lakshman , don't think false required here as entry-uid flag is of string type. pls handle it properly like this
if(!entriesFlags.entryUid){
//logic
}

packages/contentstack-utilities/package.json
```diff
@@ -32,7 +32,7 @@
   "author": "contentstack",
   "license": "MIT",
   "dependencies": {
-    "@contentstack/management": "~1.17.0",
+    "@contentstack/management": "~1.18.0",
```
aman19K, line 35: @sunil-lakshman , Please update the management SDK version anywhere the cli is used, such as the contentstack root folder or import.

packages/contentstack-bulk-publish/src/consumer/publish.js
sunil-lakshman: @aman19K Passing api_version in header is removed and set api_version in header handed in SDK side.
CC: @shafeeqd959 @cs-raj

aman19K (approved):

PR #1598, opened 2024-09-30
feat: oauth support in variant & personalize

aman19K:
- [x] OAuth support in variant & personalize

### DIFF
--- a/packages/contentstack-utilities/src/authentication-handler.ts
+++ b/packages/contentstack-utilities/src/authentication-handler.ts
@@ -1,0 +1,127 @@
+import { cliux as ux, authHandler, configHandler } from './index';
+
+class AuthenticationHandler {
+  private authType: string;
+  private isOAuth: boolean;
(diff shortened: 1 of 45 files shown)


packages/contentstack-utilities/src/authentication-handler.ts
```diff
@@ -0,0 +1,127 @@
...
+            if (region.cma.startsWith('http')) {
+              const u = new URL(region.cma);
+              if (u.host) hostName = u.host;
+            }
+            hostName = hostName || region.cma;
+            await this.refreshToken(hostName);
+            return this.refreshAccessToken(error, maxRetryCount); // Retry after refreshing the token
+          }
+
+        case 429:
```
sunil-lakshman, line 81: @aman19K Here need to handle Rate Limit Exceeded error right?

sunil-lakshman (approved):
Approved

sunil-lakshman (approved):

PR #1663, opened 2024-11-06
asset mapper, marketplace

shafeeqd959:


packages/contentstack-import-setup/src/import/import-setup.ts
```diff
@@ -81,6 +81,12 @@ export default class ImportSetup {
    */
   async start() {
     try {
+      if (!this.config.management_token) {
```
sunil-lakshman, line 84: @shafeeqd959 Can you use the Nullish Coalescing. It ensures that the values are only assigned if they are not null or undefined.
if (!this.config.management_token) {
  const { name, org_uid } = await this.stackAPIClient.fetch();
  this.config.stackName = name ?? this.config.stackName;
  this.config.org_uid = org_uid ?? this.config.org_uid;
}

cs-raj (approved):

packages/contentstack-import-setup/src/import/import-setup.ts
shafeeqd959: since his.config.stackName is already null so no need of updating this

PR #1717, opened 2025-01-03
Implemented nested global fields

sunil-lakshman:


packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -140,17 +140,18 @@ export default class ImportGlobalFields extends BaseClass {
       if (flag.supressed) {
         this.pendingGFs.push(globalField.uid);
       }
+      const hasNested = this.hasNestedGlobalFields(globalField as GlobalFieldData);
       return this.stack
-        .globalField()
+        .globalField(null, hasNested ? { api_version: nestedGlobalFieldsVersion} : undefined)
         .create({ global_field: globalField as GlobalFieldData })
```
shafeeqd959, line 146: If global field has nested references, and the referenced global field is not created yet, then how this implementation going to work ?


CC: @aman19K

shafeeqd959 (approved):

packages/contentstack-import/src/import/modules/global-fields.ts
aman19K: This implementation isn't carried out as discussed and mentioned on the Confluence Page, @shafeeqd959 .
I've also told @sunil-lakshman  that this implementation is incorrect and that he needs to change it.

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -160,6 +185,50 @@ export default class ImportGlobalFields extends BaseClass {
     });
   }
 
+  async updateGFs(): Promise<any> {
+    this.pendingGFs = fsUtil.readFile(this.gFsPendingPath) as any;
+    this.gFs = fsUtil.readFile(path.resolve(this.gFsFolderPath, this.gFsConfig.fileName)) as Record<string, unknown>[];
```
aman19K, line 190: @sunil-lakshman , Only the pending GF is being updated at this time; you need to iterate and skip the pending GF update.

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -137,20 +159,23 @@ export default class ImportGlobalFields extends BaseClass {
       lookupExtension(this.config, globalField.schema, this.config.preserveStackVersion, this.installedExtensions);
       let flag = { supressed: false };
       await removeReferenceFields(globalField.schema, flag, this.stack);
+      if (Array.isArray(globalField.schema)) {
+        await this.removeNestedGlobalFields(globalField.schema, flag);
+      }
```
aman19K, line 164: @sunil-lakshman , this code isn't required. Create GF with schema template like this https://github.com/contentstack/cli/blob/main/packages/contentstack-import/src/import/modules/content-types.ts#L172

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -137,20 +159,23 @@ export default class ImportGlobalFields extends BaseClass {
       lookupExtension(this.config, globalField.schema, this.config.preserveStackVersion, this.installedExtensions);
```
aman19K, line 159: @sunil-lakshman , look up extension not required here as we're handling this condition in update GF function.

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -137,20 +159,23 @@ export default class ImportGlobalFields extends BaseClass {
...
       await removeReferenceFields(globalField.schema, flag, this.stack);
+      if (Array.isArray(globalField.schema)) {
+        await this.removeNestedGlobalFields(globalField.schema, flag);
+      }
       if (flag.supressed) {
         this.pendingGFs.push(globalField.uid);
       }
       return this.stack
-        .globalField()
+        .globalField(null, { api_version: nestedGlobalFieldsVersion})
```
aman19K, line 169: @sunil-lakshman , do we need to provide null here ?

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -124,6 +128,24 @@ export default class ImportGlobalFields extends BaseClass {
...
+        schema.splice(i, 1);
+        flag.supressed = true;
+      } else if (field.schema) {
+        // Recursively check nested schemas
+        await this.removeNestedGlobalFields(field.schema, flag);
+      }
+    }
+  }
+  
+
```
aman19K, line 148: @sunil-lakshman , As we're seeding all global fields with schemaTemplate similar like CT then this isn't required.

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -85,8 +89,8 @@ export default class ImportGlobalFields extends BaseClass {
     log(this.config, 'Global fields import has been completed!', 'info');
   }
 
-  async importGFs() {
-    const onSuccess = ({ response: globalField, apiData: { uid } = undefined }: any) => {
+  async seedGFs() {
+    const onSuccess = async ({ response: globalField, apiData: { uid } = undefined }: any) => {
```
aman19K, line 93: @sunil-lakshman , do we need async here ?

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -160,6 +185,50 @@ export default class ImportGlobalFields extends BaseClass {
...
+      },
+      concurrencyLimit: this.reqConcurrency,
+    });
+  }
+
+  serializeUpdateGFs(apiOptions: ApiOptions): ApiOptions {
+    const {
+      apiData: { uid },
+    } = apiOptions;
+    const globalField = find(this.gFs, { uid });
```
aman19K, line 217: @sunil-lakshman ,You can provide the entire payload in place of the UID. pls take ref from CT update https://github.com/contentstack/cli/blob/main/packages/contentstack-import/src/import/modules/content-types.ts#L206

packages/contentstack-import/src/import/modules/global-fields.ts
sunil-lakshman: @aman19K we need to remove the schema template have data-type "global fields", Otherwise we will get below error if reference_to value is not present in the global fields.
```
{
    "error_message": "Global Field creation failed. Please try again.",
    "error_code": 115,
    "errors": {
        "schema.1.reference_to": [
            "123 refers to a Global Field that does not exist."
        ]
    }
}
```

packages/contentstack-import/src/import/modules/global-fields.ts
sunil-lakshman: @aman19K  https://github.com/contentstack/cli/pull/1717/files#r1906466098

packages/contentstack-import/src/import/modules/global-fields.ts
sunil-lakshman: @aman19K The current implementation is intentional, as the logic is designed to update only the pending global fields. These pending global fields are identified during the seedGFs step because they contain nested global fields. Updating only the pending ones ensures that we handle the specific cases where nested global fields exist, without unnecessarily processing other global fields that do not have nested dependencies.

packages/contentstack-import/src/import/modules/global-fields.ts
sunil-lakshman: @aman19K In CT update method also passing only the UID.
https://github.com/contentstack/cli/blob/b51eeae2600d070495d8144cf32f3dc4d3d0bcae/packages/contentstack-import/src/import/modules/content-types.ts#L262

packages/contentstack-import/src/import/modules/global-fields.ts
aman19K: @sunil-lakshman , that's why we've to create GF with this template
{
  "title": "Seed",
  "uid": "",
  "schema": [
      {
        "display_name": "Title",
        "uid": "title",
        "data_type": "text",
        "field_metadata": {
          "_default": true
        },
        "unique": false,
        "mandatory": true,
        "multiple": false
      }
  ],
  "description": "Seed "
}

packages/contentstack-import/src/import/modules/global-fields.ts
aman19K: @sunil-lakshman , As discussed, pls fix this.

packages/contentstack-import/src/import/modules/global-fields.ts
aman19K: @sunil-lakshman , As discussed, pls fix this.

packages/contentstack-import/src/import/modules/global-fields.ts
aman19K: @sunil-lakshman , As discussed, pls fix this.

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -85,79 +91,96 @@ export default class ImportGlobalFields extends BaseClass {
...
+    removeReferenceFields(globalField.schema, flag, this.stack);
+    if (flag.supressed) {
+      this.pendingGFs.push(globalField.uid);
+    }
+    const globalFieldPayload = this.stack.globalField(globalField.uid, { api_version: '3.2' });
+    Object.assign(globalFieldPayload, cloneDeep(globalField), {
+      stackHeaders: globalFieldPayload.stackHeaders,
     });
+    apiOptions.apiData = globalFieldPayload;
+    return apiOptions;
```
aman19K, line 183: @sunil-lakshman , In order to bypass the update for pending GF, we must provide entity undefined in this case. Like this
  ```
  if (flag.supressed) {
      this.pendingGFs.push(globalField.uid);
      apiOptions.entity = undefined;
    }else{
           const globalFieldPayload = this.stack.globalField(globalField.uid, { api_version: '3.2' });
        Object.assign(globalFieldPayload, cloneDeep(globalField), {
          stackHeaders: globalFieldPayload.stackHeaders,
        });
       apiOptions.apiData = globalFieldPayload;
    }
    return apiOptions;
```

**Comment by aman19K** [comment]
@sunil-lakshman , LGTM.

**Review by aman19K — APPROVED** [review]

File: packages/contentstack-import/src/import/modules/base-class.ts [hunk 10]
```diff
@@ -315,6 +315,8 @@ export default abstract class BaseClass {
         return this.stack.contentType().create(apiData).then(onSuccess).catch(onReject);
       case 'update-cts':
         return apiData.update().then(onSuccess).catch(onReject);
+      case 'create-gfs':
+        return this.stack.globalField({api_version: '3.2'}).create(apiData).then(onSuccess).catch(onReject);
```
shafeeqd959, line 319: Global api call is always going to be 3.2, I think not all customers will be having 3.2 api access. could check this,  if so how are we going to manage this ?

sunil-lakshman: @shafeeqd959 We can provide version 3.2 by default. They have added a check at the service level to verify if the version is 3.2 and if there are nested global fields, in such cases, it routes to the new service, and if not, it routes to the old service.
I have checked, and it's working fine as expected.

CC: @aman19K

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -85,81 +91,101 @@ export default class ImportGlobalFields extends BaseClass {
...
+      concurrencyLimit: this.reqConcurrency,
     });
   }
 
+  /**
+   * @method serializeUpdateGFs
+   * @param {ApiOptions} apiOptions ApiOptions
+   * @returns {ApiOptions} ApiOptions
+   */
+  async serializeUpdateGFs(apiOptions: ApiOptions): Promise<ApiOptions> {
```
shafeeqd959, line 170: async serializeUpdateGFs can't be provided as a serialize function

aman19K (approved):

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -160,6 +200,21 @@ export default class ImportGlobalFields extends BaseClass {
...
```
shafeeqd959, line 208: Why do we need this "serializeUpdateGFs" function are we using it anywhere ? i see this is being send to make concurrent call,  how that is useful 

CC: @aman19K @sunil-lakshman

packages/contentstack-import/src/import/modules/global-fields.ts
```diff
@@ -85,46 +91,80 @@ export default class ImportGlobalFields extends BaseClass {
...
+    };
+    
+    return await this.makeConcurrentCall({
+      processName: 'Update Global Fields',
+      apiContent: this.gFs,
+      apiParams: {
+        serializeData: this.serializeUpdateGFs.bind(this),
+        reject: onReject.bind(this),
+        resolve: onSuccess.bind(this),
+        entity: 'update-gfs',
```
shafeeqd959, line 158: Why do we still need to have serialize data function and an entity here ? since we have a promisifying handler

packages/contentstack-import/src/import/modules/base-class.ts
```diff
@@ -315,6 +315,8 @@ export default abstract class BaseClass {
         return this.stack.contentType().create(apiData).then(onSuccess).catch(onReject);
       case 'update-cts':
         return apiData.update().then(onSuccess).catch(onReject);
+      case 'create-gfs':
+        return this.stack.globalField({api_version: '3.2'}).create(apiData).then(onSuccess).catch(onReject); 
       case 'update-gfs':
```
shafeeqd959, line 320: We already have global field update method in 'updateGFs' here we kept the implementation for the update as well, in two places

packages/contentstack-import/src/import/modules/content-types.ts
```diff
@@ -13,6 +13,12 @@ import BaseClass, { ApiOptions } from './base-class';
 import { updateFieldRules } from '../../utils/content-type-helper';
 import { sanitizePath } from '@contentstack/cli-utilities';
 
+type GlobalFieldData = {
```
shafeeqd959, line 16: can you move this type definition to the specific folder where we keep types

sunil-lakshman: @shafeeqd959 The code block has been removed, as it is no longer being used.

packages/contentstack-import/src/import/modules/base-class.ts
sunil-lakshman: @shafeeqd959 We are using the "update-gfs" method in the ContentTypesImport class to update the pending global fields.

CC: @aman19K

packages/contentstack-import/src/import/modules/global-fields.ts
sunil-lakshman: @shafeeqd959 Removed the "serializeUpdateGFs" method

PR #1724, opened 2025-01-13
Feat/dx 1990

shafeeqd959:


packages/contentstack-import-setup/src/utils/common-helper.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const validateBranch = async (stackAPIClient: any, config: ImportConfig, branch: any) => {
+  return new Promise(async (resolve, reject) => {
+    try {
+      const data = await stackAPIClient.branch(branch).fetch();
+      if (data && typeof data === 'object') {
+        if (data.error_message) {
+          log(config, chalk.red(data.error_message), 'error');
+          log(config, chalk.red('No branch found with the name ' + branch), 'error');
+          reject({ message: 'No branch found with the name ' + branch, error: data.error_message });
```
aman19K, line 13: @shafeeqd959 , [Minor] we can keep these error messages [here](https://github.com/contentstack/contentstack-apps-cli/blob/main/src/messages/index.ts). It'll easy to manage.

aman19K (approved):

packages/contentstack-import-setup/src/commands/cm/stacks/import-setup.ts
```diff
@@ -29,15 +33,23 @@ export default class ImportSetupCommand extends Command {
...
+      options: ['global-fields', 'content-types', 'entries'], // only allow the value to be from a discrete set
+      description:
+        '[optional] Specify the modules/module to import into the target stack. currently options are global-fields, content-types, entries',
+      multiple: true,
+    }),
+    branch: flags.string({
+      char: 'B',
+      description:
+        "The name of the branch where you want to import your content. If you don't mention the branch name, then by default the content will be imported to the main branch.",
+      parse: printFlagDeprecation(['-B'], ['--branch']),
```
aman19K, line 52: @shafeeqd959 , do we need char and printFlagDeprecation here? As it's new plugin.

packages/contentstack-import-setup/src/commands/cm/stacks/import-setup.ts
```diff
@@ -16,10 +16,14 @@ import { setupImportConfig, log } from '../../../utils';
...
+  static description = messageHandler.parse(
+    'Helps to generate mappers and backup folder for importing (overwriting) specific modules',
+  );
 
   static examples: string[] = [
-    `csdx cm:stacks:import-setup --stack-api-key <stack_api_key> --data-dir <path/of/export/destination/dir> --modules <module_name, module_name>`,
+    `csdx cm:stacks:import-setup --stack-api-key <target_stack_api_key> --data-dir <path/of/export/destination/dir> --modules <module_name, module_name>`,
+    `csdx cm:stacks:import-setup -k <target_stack_api_key> -d <path/of/export/destination/dir> --modules <module_name, module_name>`,
+    `csdx cm:stacks:import-setup -k <target_stack_api_key> -d <path/of/export/destination/dir> --modules <module_name, module_name> -b <branch_name>`,
   ];
```
aman19K, line 28: @shafeeqd959 , can you pls use long flag in the example?

aman19K (approved):

packages/contentstack-import-setup/src/commands/cm/stacks/import-setup.ts
shafeeqd959: @aman19K , currently doc is updated with same details, will remove it in the upcoming releases

packages/contentstack-import-setup/src/commands/cm/stacks/import-setup.ts
shafeeqd959: its already added in the first place

PR #1725, opened 2025-01-15
fix: Sanitized the path to avoid excluding the files

cs-raj:
version bump:
@contentstack/cli-utilities: 1.8.0 -> 1.8.1


aman19K (dismissed):

shafeeqd959:
Need version bumps in export, import, import-setup and migrate rte, then content stack package.json

PR #1743, opened 2025-02-03
fix: Removed the unused dependency from the cli package

cs-raj:


shafeeqd959:
Versions of each package needs to be updated, then contentstack package.json as well

cs-raj:
> Versions of each package needs to be updated, then contentstack package.json as well

Except contentstack rest all package version has been bumped as it will be updated during the release preperation

aman19K (approved):

PR #1754, opened 2025-02-10
SRE fix

cs-raj:


shafeeqd959 (approved):

packages/contentstack-import/src/utils/entries-helper.ts
```diff
@@ -212,6 +213,8 @@ export const lookupEntries = function (
       if (status === 'safe') {
         entry = entry.replace(uidRegex, escapedMappedUid);
         mapped.push(uid);
+      } else {
+        log(`Skipping the entry uid ${uid} since the regex is not valid`,'error');
```
shafeeqd959, line 217: Update the log level to warning

cs-raj: Done
cc: @aman19K

PR #1756, opened 2025-02-11
[DX-71], modified condition to print detailed error in case of certificate error

naman-contentstack:


cs-raj (approved):

naman-contentstack:
> @naman-contentstack , pls do patch version bump

patch version is already updated in development

aman19K (approved):

PR #1764, opened 2025-02-12
DX - 2182 -  Added use of delivery token alias in bulk publish

cs-raj:
version bump:
1.7.2 -> 1.8.0


harshithad0703 (approved):

sunil-lakshman (approved):
@cs-raj Looks good to me.
Please take approval from @shafeeqd959 & @aman19K

PR #1767, opened 2025-02-14
semgrep issue

harshithad0703:
- refactor: update uid validation to use static regex and switch to HTTPS server

### DIFF
--- a/packages/contentstack-import/src/utils/entries-helper.ts
+++ b/packages/contentstack-import/src/utils/entries-helper.ts
@@ -206,11 +206,10 @@ export const lookupEntries = function (
   let entry = JSON.stringify(data.entry);
   uids?.forEach(function (uid: any) {
     if (mappedUids.hasOwnProperty(uid)) {
-      const sanitizedUid = escapeRegExp(uid);
       const escapedMappedUid = escapeRegExp(mappedUids[uid]);
-      const uidRegex = new RegExp(`\\b${sanitizedUid}\\b`, 'img');
-      let { status } = validateRegex(uidRegex);
-      if (status === 'safe') {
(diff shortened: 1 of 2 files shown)


packages/contentstack-utilities/src/auth-handler.ts
```diff
@@ -119,7 +120,7 @@ class AuthHandler {
   async createHTTPServer(): Promise<object> {
     return new Promise((resolve, reject) => {
       try {
-        const server = http.createServer((req, res) => {
+        const server = https.createServer((req, res) => {
```
aman19K, line 123: @harshithad0703 ,We can't use HTTPS in the CLI because the server is hosted on the client side, so it's not possible. Pls check with @cs-raj regarding this.

packages/contentstack-import/src/utils/entries-helper.ts
```diff
@@ -206,11 +206,10 @@ export const lookupEntries = function (
...
   uids?.forEach(function (uid: any) {
     if (mappedUids.hasOwnProperty(uid)) {
-      const sanitizedUid = escapeRegExp(uid);
       const escapedMappedUid = escapeRegExp(mappedUids[uid]);
-      const uidRegex = new RegExp(`\\b${sanitizedUid}\\b`, 'img');
-      let { status } = validateRegex(uidRegex);
-      if (status === 'safe') {
+      // Use a static regex for alphanumeric only uids
+      const uidRegex = /^[a-zA-Z0-9]+$/; // Static regex that matches alphanumeric uids
+      if (uid.match(uidRegex)) {  // Validate against the static regex
```
aman19K, line 212: @harshithad0703 , pls check with @cs-raj. Both issues we can't fix.

PR #1772, opened 2025-02-18
Feature nested global fields

sunil-lakshman:


packages/contentstack-export/src/export/modules/content-types.ts
```diff
@@ -68,6 +68,7 @@ export default class ContentTypesExport extends BaseClass {
     if (skip) {
       this.qs.skip = skip;
     }
+    delete this.stackAPIClient.stackHeaders.api_version;
```
aman19K, line 71: @sunil-lakshman , could you please check if we still require to delete api_version from headers, as I remember it started due to a management SDK bug?

cs-raj (approved):

aman19K (approved):

PR #1791, opened 2025-03-03
test: added custom roles test cases

vkalta:


aman19K:
@vkalta , LGTM, but as the modifications are independent of nested global fields, please modify the base branch.

PR #1812, opened 2025-03-15
DX - 2324 - Feat: Added Support for audit fix to fix select field based on config and added fixedStatus and selected value

cs-raj:
version bump: 1.8.0 -> 1.8.1

### DIFF
--- a/packages/contentstack-audit/src/modules/entries.ts
+++ b/packages/contentstack-audit/src/modules/entries.ts
@@ -854,7 +854,7 @@ export default class Entries {
 
     let missingCTSelectFieldValues;
     let isMissingValuePresent = false;
-    let selectedValue: string | string [] | number | number[] | any = '';
+    let selectedValue: unknown = '';
     if (multiple) {
       let obj = this.findNotPresentSelectField(entry, selectOptions);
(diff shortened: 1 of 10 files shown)


packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -846,11 +847,14 @@ export default class Entries {
...
   fixSelectField(tree: Record<string, unknown>[], field: SelectFeildStruct, entry: any) {
+    if(!this.config.fixSelectField) {
+      return entry;
+    }
     const { enum: selectOptions, multiple, min_instance, display_type, display_name, uid } = field;
 
     let missingCTSelectFieldValues;
     let isMissingValuePresent = false;
-
+    let selectedValue: string | string [] | number | number[] | any = '';
```
aman19K, line 857: @cs-raj , use **unknown** instead of **any**. As it allows any type of value to be assigned.

cs-raj: done @aman19K

aman19K:
@cs-raj , LGTM

harshithad0703 (approved):

PR #1813, opened 2025-03-17
Check for the validation of mandatory fields after the fix

cs-raj:


sunil-lakshman (approved):
LGTM

cs-raj:
Hi Team,
Closing this PR since if a field is mandatory and it does not contain any data it will be removed already so the check for mandatory field after the field fixing will add no benefit
cc: @netrajpatel @shafeeqd959 @aman1908

PR #1816, opened 2025-03-17
fix: handling ignored content-types

vkalta:


shafeeqd959:
"If the user presses the Enter key instead of the space bar, no item will be selected for merge.
Need to display proper error message." - this needs to be handled

packages/contentstack-branches/src/interfaces/index.ts
```diff
@@ -13,6 +13,7 @@ export interface BranchDiffRes {
   title: string;
   type: string;
   status: string;
+  merge_strategy?: string;
 }
 
 export interface BranchDiffSummary {
```
shafeeqd959, line 19: bump versions

shafeeqd959 (approved):

aman19K (approved):

shafeeqd959 (approved):

PR #1827, opened 2025-03-23
fix: added missing branch flag handling

vkalta:


packages/contentstack-import/src/commands/cm/stacks/import.ts
```diff
@@ -148,6 +148,27 @@ export default class ImportCommand extends Command {
...
+          if (branches.length) {
+            flags.branch = 'main';
+          }
+        } catch (error) {
+          // Branch not enabled, just the let flow continue
+          log(
+            importConfig,
+            `The branch is not enabled for this stack - ${formatError(error)}`,
+            'error',
+          );
```
aman19K, line 169: @vkalta, we don't need to log error message in case branch isn't enabled. In this case catch block should be empty

vkalta: @aman19K This is now done.

aman19K (approved):

cs-raj (approved):

PR #1835, opened 2025-04-01
feat: Add workflow to run and validate unit tests

aman19K:
- [x] Add workflow to run and validate unit tests
- [x] version bump already done

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -23,7 +23,3 @@ jobs:
       - name: Run tests for Audit plugin
         working-directory: ./packages/contentstack-audit
         run: npm run test:unit
-      
-      - name: Run tests for Export to CSV plugin
-        working-directory: ./packages/contentstack-export-to-csv
-        run: npm run test:unit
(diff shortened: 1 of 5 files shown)


cs-raj (approved):

.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,25 @@
...
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22.x'
+
+      - name: Install dependencies for all plugins
+        run: |
+          npm run setup-repo-old
```
shafeeqd959, line 21: Is possible to update the workflow in a way that it unit tests runs only the plugins which has updates or changes, Unlike release pipeline, this will be executed on every PR, this workflow will eventually executes unit tests of all plugins, Is it really necessary ?

aman19K: @shafeeqd959 done

shafeeqd959 (approved):

PR #1844, opened 2025-04-07
[DX-2346], added support for AWS-NA and AWS-EU, fixed test cases

naman-contentstack:


packages/contentstack-config/src/utils/interactive.ts
```diff
@@ -9,12 +9,14 @@ export const askRegions = async (): Promise<string> => {
     choices: [
       { name: 'NA', value: 'NA' },
       { name: 'EU', value: 'EU' },
+      { name: 'AWS-NA', value: 'AWS-NA' },
+      { name: 'AWS-EU', value: 'AWS-EU' },
```
aman19K, line 13: @naman-contentstack , pls check I think in case of AWS-NA and AWS-EU the value will be also NA and EU.
CC:  @netrajpatel

aman19K:
@naman-contentstack , pls update the branch and also if all config test cases are working then please add the check in github workflow.

aman19K (approved):

PR #1846, opened 2025-04-07
[DX-2288], upgraded recheck version

naman-contentstack:


naman-contentstack:
created a new PR, due to many conflicts in package file.
https://github.com/contentstack/cli/pull/1896

PR #1886, opened 2025-04-23
fix: Handle empty object in payload causing 400 Bad Request in GCP region

aman19K:
**Description**
This PR addresses an issue where sending an empty object in the payload results in a 400 Bad Request error. The backend was unable to process the request due to the malformed structure, which returned a generic HTML error response.

**Changes Made**
- [x] Validation added to prevent sending an empty object in the request payload.
- [x] Error handling improved to catch and display  complete error when this kind of response is received from the server.
- [x] version bump


sunil-lakshman (dismissed):
LGTM

cs-raj (dismissed):

sunil-lakshman (approved):
LGTM

PR #1895, opened 2025-04-25
test: fix failing cases and added check in workflow

aman19K:
- [x] fix failing cases and added check in workflow
- [x] skipped two cases because it's taking more than 30-40 sec.
- [x] version bump


sunil-lakshman (approved):
LGTM

cs-raj (approved):

PR #1940, opened 2025-05-28
Fixed Nested Global Field

cs-raj:


packages/contentstack-import/src/utils/content-type-helper.ts
```diff
@@ -86,6 +86,7 @@ export const removeReferenceFields = async function (
   flag = { supressed: false },
   stackAPIClient: any,
 ): Promise<boolean | void> {
+  if(schema && schema.length){
```
aman19K, line 89: @cs-raj , it's better use optional chaining here if(schema?.length)

aman19K (approved):

PR #1955, opened 2025-06-09
Fix/dx 2953 export import stack settings

harshithad0703:


packages/contentstack-export/src/export/modules/stack.ts
```diff
@@ -29,11 +29,18 @@ export default class ExportStack extends BaseClass {
         this.exportConfig.sourceStackName = stackData.name;
       }
     }
-    if (!this.exportConfig.preserveStackVersion && !this.exportConfig.hasOwnProperty('master_locale')) {
-      //fetch master locale details
+
+    // Always export stack settings first
+    await this.exportStackSettings();
```
aman19K, line 34: @harshithad0703 , exportStackSettings should be inside authenticated block.

harshithad0703: ok

PR #1959, opened 2025-06-09
Added AWS-AU region support in config command

netrajpatel:


cs-raj (approved):

aman19K (dismissed):

tsconfig.json
```diff
@@ -0,0 +1,103 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
+
```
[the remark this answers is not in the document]
netrajpatel: @shafeeqd959 Without this file, some packages are giving an error that they cannot find tsconfig.json; hence, it has been added.

packages/contentstack-import/src/import/modules/entries.ts
```diff
@@ -817,19 +823,17 @@ export default class EntriesImport extends BaseClass {
   async removeAutoCreatedEntries(): Promise<void> {
     const onSuccess = ({ response, apiData: { entryUid } }: any) => {
       // NOTE Remove entry from list
-      remove(this.entriesForVariant, {
-        entry_uid: entryUid,
-        locale: this.importConfig?.master_locale?.code,
-      });
+      this.entriesForVariant = this.entriesForVariant.filter(
```
shafeeqd959, line 826: Why was this code updated to inline logic instead of using a method? This change has introduced redundancy across the codebase

packages/contentstack/package.json
```diff
@@ -28,14 +28,14 @@
     "@contentstack/cli-cm-branches": "~1.4.3",
     "@contentstack/cli-cm-bulk-publish": "~1.8.1",
     "@contentstack/cli-cm-clone": "~1.14.2",
-    "@contentstack/cli-cm-export": "~1.16.2",
+    "@contentstack/cli-cm-export": "~1.16.3",
```
shafeeqd959, line 31: The export and import changes should be treated as a minor version update, as the code modifications introduce updated region naming conventions rather than fixing any bugs.

netrajpatel: This is resolved

packages/contentstack-import/src/import/modules/entries.ts
netrajpatel: Since this was an existing implementation, I just resolved the type issues in all places.

PR #1995, opened 2025-07-11
feat: integrated logger in variant(import)

aman19K:


packages/contentstack-import/src/commands/cm/stacks/import.ts
```diff
@@ -16,7 +15,7 @@ import {
 
 import { Context, ImportConfig } from '../../../types';
 import { ModuleImporter } from '../../../import';
-import { setupImportConfig } from '../../../utils';
+import { setupImportConfig,  } from '../../../utils';
```
sunil-lakshman, line 18: @aman19K Unnecessary trailing comma — can be removed for clarity.

cs-raj (approved):

sunil-lakshman (approved):
LGTM

PR #2045, opened 2025-08-05
DX - 3258 -  Added Variant support in Audit plugin to check for publish_details and references 

cs-raj:
version bump: 1.14.0 -> 1.15.0

### DIFF
--- a/packages/contentstack-audit/src/audit-base-command.ts
+++ b/packages/contentstack-audit/src/audit-base-command.ts
@@ -31,6 +31,7 @@ import {
   OutputColumn,
   RefErrorReturnType,
   WorkflowExtensionsRefErrorReturnType,
+  ScanAndFixResult,
 } from './types';
 
 export abstract class AuditBaseCommand extends BaseCommand<typeof AuditBaseCommand> {
@@ -64,6 +65,7 @@ export abstract class AuditBaseCommand extends BaseCommand<typeof AuditBaseComma
       missingCtRefs,
       missingGfRefs,
       missingEntryRefs,
+      missingVariantRefs,
       missingCtRefsInExtensions,
       missingCtRefsInWorkflow,
       missingSelectFeild,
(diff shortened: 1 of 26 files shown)


packages/contentstack-audit/src/audit-base-command.ts
```diff
@@ -568,6 +576,11 @@ export abstract class AuditBaseCommand extends BaseCommand<typeof AuditBaseComma
             row['Fix status'] = row.fixStatus;
           }
 
+          // Handle variant flag for entries
+          if ('is_variant' in issue) {
```
aman19K, line 580: @cs-raj , can we add is_variant in constant files or somewhere else? So that we can use the variable in test cases as well.

packages/contentstack-audit/src/modules/entries.ts
```diff
@@ -1,4 +1,7 @@
-import map from 'lodash/map';
```
aman19K, line 1: @cs-raj , try to use import map from 'lodash/map';

packages/contentstack-audit/src/modules/variant/variant-logger.ts
```diff
@@ -0,0 +1,15 @@
...
+  constructor(private readonly log: LogFn) {}
+
+  logError(message: string, args: Record<string, string>): void {
+    this.log($t(message, args), 'error');
+  }
+
+  logInfo(message: string, args: Record<string, string>): void {
+    this.log($t(message, args), 'info');
+  }
+}
```
aman19K, line 15: @cs-raj , why have you added separate logger for variant? We can use existing audit log file

packages/contentstack-audit/src/modules/variant/variant-publish-validator.ts
```diff
@@ -0,0 +1,236 @@
...
+    private readonly fix: boolean = false,
+    entryMetaData: Record<string, any>[] = []
+  ) {
+    this.referenceValidator = new VariantReferenceValidator(entryMetaData, logger, fix);
+  }
+
+  private isValidVariantEntry(entry: unknown): entry is VariantEntry {
+    if (!entry || typeof entry !== 'object') return false;
+    
+    const e = entry as Record<string, unknown>;
```
aman19K, line 25: @cs-raj , instead of e variable can you use meaningful name ?

packages/contentstack-audit/src/modules/variant/variant-reference-validator.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    // Get all values from change set paths
+    const changeSetValues = this.accessChangeSetFields(variant, variant._variant._change_set);
+
+    // First validate all references in the variant object against entryMetadata
+    for (const { path, value } of changeSetValues) {
+      if (Array.isArray(value)) {
+        // Check each item in array for references
+        for (const item of value) {
+          if (item && typeof item === 'object' && item.uid && item._content_type_uid) {
```
aman19K, line 29: @cs-raj , use optional chaining here

aman19K (changes requested):

cs-raj:
Closing the PR

PR #2046, opened 2025-08-06
feat: Integrated CLIProgressManager and SummaryManager in variant export

aman19K:


packages/contentstack-export/src/utils/marketplace-app-helper.ts
```diff
@@ -35,6 +38,7 @@ export async function createNodeCryptoInstance(config: ExportConfig): Promise<No
       },
       message: 'Enter Marketplace app configurations encryption key',
     });
+    console.log('');
```
naman-contentstack, line 41: here aswell, can we use cliux instead of console.logs?

naman-contentstack (approved):

PR #2048, opened 2025-08-07
Fixed branches plugin unit testcases

sunil-lakshman:


.talismanrc
```diff
@@ -0,0 +1,4 @@
+fileignoreconfig:
+- filename: package-lock.json
+  checksum: 04099ea4a522a2be7c5b0c3a3c010a6feab6b2a078ef6d1145450953f63ac0de
+version: ""
```
aman19K, line 4: @sunil-lakshman , why talismanrc file contains only package-lock.json? In development branch I've noticed that lots of file we're skipping pls validate it.

aman19K (approved):

.talismanrc
netrajpatel: This is expected; let's resolve this.

PR #2057, opened 2025-08-11
Enable tests for Contentstack Import Setup in unit-test.yml

harshithad0703:


harshitha07d:
The test cases were fixed when cli was modularized
https://github.com/contentstack/cli-plugins/pull/5

PR #2082, opened 2025-08-22
Fix/config test cases

harshithad0703:


shafeeqd959:
@sunil-lakshman mention ticket number

PR #2084, opened 2025-08-22
Fix/add bulk url

harshithad0703:
dx-2196 Added bulk publish queue link


packages/contentstack-bulk-publish/src/util/bulk-publish-utils.js
```diff
@@ -0,0 +1,39 @@
+const { regions } = require('@contentstack/cli-config/lib/utils/region-handler');
```
shafeeqd959, line 1: can you name this file better, since the folder tells it that is a group of utilities, give name specific to its usage for better understanding such as "generate-bulk-publish-url.js"

packages/contentstack-bulk-publish/package.json
```diff
@@ -1,11 +1,12 @@
 {
   "name": "@contentstack/cli-cm-bulk-publish",
   "description": "Contentstack CLI plugin for bulk publish actions",
-  "version": "1.9.0",
+  "version": "1.9.1",
```
shafeeqd959, line 4: @sunil-lakshman its a new feature not a bug fix or a patch, for feature updates we have to bump the minor version,  this should reflects in the contentstack package.json as well cli will get minor update

shafeeqd959:
@sunil-lakshman please mention the ticket number like fix/CS-123434, also please get @AniketDev7 tested this before merge it to development

aman19K (approved):

PR #2122, opened 2025-09-16
update logger to v2 and fix import path validation

naman-contentstack:


packages/contentstack-import/src/import/module-importer.ts
```diff
@@ -69,7 +65,7 @@ class ModuleImporter {
           this.importConfig.moduleName,
         ))
     ) {
-      if (!(await this.auditImportData(logger))) {
+      if (!(await this.auditImportData())) {
```
aman19K, line 68: @naman-contentstack , can you check I think we're using this logger in audit?

packages/contentstack-import/src/import/module-importer.ts
```diff
@@ -30,14 +29,14 @@ class ModuleImporter {
...
+    await this.resolveImportPath();
+
     if (!this.importConfig.management_token) {
       const stackDetails: Record<string, unknown> = await this.stackAPIClient.fetch();
       this.importConfig.stackName = stackDetails.name as string;
       this.importConfig.org_uid = stackDetails.org_uid as string;
     }
 
-    await this.resolveImportPath();
-
```
aman19K, line 40: @naman-contentstack , why have you changed the seq?

aman19K (approved):

PR #2154, opened 2025-10-07
fix: export clearProgressModuleSetting function from utilities index

aman19K:
Fixes: TypeError: (0 , cli_utilities_1.clearProgressModuleSetting) is not a function
Affects: cm:stacks:export and cm:stacks:import commands in CLI beta

RCA: Utilities plugin not released


AniketDev7 (approved):
LGTM

PR #2200, opened 2025-10-29
feat: update region handling and add regions.json for dynamic region …

nadeem-cs:
…configuration


harshitha07d:
https://github.com/contentstack/cli/pull/2377
The endpoint integration is completed in this PR

PR #2239, opened 2025-11-14
Feat/dx 3725

netrajpatel:


aman19K:
@netrajpatel , Since we're planning to separate out migrate plugin from CLI core,  do we need to to remove deprecated flags?

netrajpatel:
> @netrajpatel , Since we're planning to separate out migrate plugin from CLI core, do we need to to remove deprecated flags?

We will clean up the deprecated stuff as 2.x-beta, and then we can separate this plugin out as 2.x so that it will not break the existing CLI.

aman19K (approved):

PR #2254, opened 2025-11-26
integrated centralized logger in import setup

sunil-lakshman:


packages/contentstack-import-setup/src/commands/cm/stacks/import-setup.ts
```diff
@@ -67,29 +70,41 @@ export default class ImportSetupCommand extends Command {
...
-      );
+      handleAndLogError(error);
     }
   }
+
+  // Create import setup context object
+  private createImportSetupContext(apiKey: string, authenticationMethod?: string): Context {
+    return {
+      command: this.context?.info?.command || 'cm:stacks:import-setup',
+      module: '',
```
naman-contentstack: why is module hardcoded empty?

packages/contentstack-import-setup/src/utils/context-helper.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export function initializeContext(
+  config: ImportConfig,
+  module?: Modules,
+  commandContext?: { info?: { command?: string }; sessionId?: string },
+): void {
+  if (!config.context) {
+    config.context = {
+      command: commandContext?.info?.command || 'cm:stacks:import-setup',
+      module: module || '',
```
naman-contentstack: only the module name will be updated by this functions, rest all of the data will be unchanged, we dont need a function for this we can set directly where ever required

naman-contentstack (commented):
please check comments

aman19K (approved):

PR #2294, opened 2026-01-06
Fix: DX-2943 validate empty api keys

nadeem-cs:


packages/contentstack-import-setup/src/utils/import-config-handler.ts
```diff
@@ -63,6 +63,9 @@ const setupConfig = async (importCmdFlags: any): Promise<ImportConfig> => {
       if (typeof config.apiKey !== 'string') {
         throw new Error('Invalid API key received');
       }
+      if (!config.apiKey || !config.apiKey.trim()) {
```
aman19K: @nadeem-cs , same here https://github.com/contentstack/cli/pull/2294/files#r2667128013

packages/contentstack-import/src/utils/import-config-handler.ts
```diff
@@ -84,6 +84,10 @@ const setupConfig = async (importCmdFlags: any): Promise<ImportConfig> => {
       if (typeof config.apiKey !== 'string') {
         throw new Error('Invalid API key received');
       }
+      if (!config.apiKey || !config.apiKey.trim()) {
```
aman19K: @nadeem-cs , same here https://github.com/contentstack/cli/pull/2294/files#r2667128013

naman-contentstack (dismissed):
LGTM

aman19K (dismissed):

aman19K:
@nadeem-cs @naman-contentstack , [Minor] moving forward try to keep the error message in message folder so that we can avoid redundant hardcoded message.
Currently if suppose someone updated the message in code and forgot to update in test cases then it'll fail.

naman-contentstack (approved):

harshitha07d:
need to do rework as we have modularized the cli and cli-plugins

PR #2319, opened 2026-01-19
feat: export and import config and code cleanup

naman-contentstack:


packages/contentstack-export/src/utils/export-config-handler.ts
```diff
@@ -24,6 +24,19 @@ const setupConfig = async (exportCmdFlags: any): Promise<ExportConfig> => {
...
+    const foundDeprecated = deprecatedProps.filter((prop) => externalConfig[prop] !== undefined);
+
+    if (foundDeprecated.length > 0) {
+      throw new Error(
+        `Unsupported configuration properties detected: ${foundDeprecated.join(', ')}. ` +
+          `These properties are no longer supported in the beta version. ` +
+          `TypeScript modules are now the default and only supported format.`,
+      );
+    }
+
```
aman19K: @naman-contentstack , Since we're releasing major version, we don't need these type of unnecessary code.

aman19K (approved):

packages/contentstack-export/src/utils/export-config-handler.ts
naman-contentstack: Removed this code

PR #2407, opened 2026-02-17
fix: updated the cli-utilites and added prepack steps for clone

cs-raj:
version bump: 
@contentstack/cli-utilities: 1.7.2 -> 1.7.3


netrajpatel (approved):
Approved

AniketDev7 (approved):

PR #2421, opened 2026-02-23
fix: log mode consistency

naman-contentstack:
## Summary
Fixes mixed console/progress output when using `--show-console-logs` and corrects the final summary label when import runs with audit (e.g. clone).

## Changes
- **Console log mode:** When `showConsoleLogs` is true, skip the global summary and per-module "Detailed Summary" so output is only timestamped log lines.
- **Clone:** Set `progressSupportedModule` to `'clone'` at the start of clone and add `'clone'` to `PROGRESS_SUPPORTED_MODULES` so logger and progress use the same mode (no mix of logs + progress bars).
- **Import after audit:** Re-initialize the global summary for IMPORT after audit so the final summary shows "IMPORT completed successfully!" instead of "AUDIT completed successfully!".

## Packages
- `cli-utilities` (progress summary, logging constants, tests)
- `cli-clone` (init)
- `cli-import` (module importer)


packages/contentstack-utilities/test/unit/cliProgressManager.test.ts
```diff
@@ -57,6 +57,7 @@ Module.prototype.require = function (id: string) {
 
 import CLIProgressManager from '../../src/progress-summary/cli-progress-manager';
 import SummaryManager from '../../src/progress-summary/summary-manager';
+import { configHandler } from '../../src';
```
aman19K: @naman-contentstack , pls check the path

aman19K (approved):

PR #2427, opened 2026-02-24
DX - 3445  Experience import errors when variants reference Lytics audiences

cs-raj:
# Fix: Experience import errors when variants reference Lytics audiences

## Problem

Personalize import fails with API errors when importing experiences whose variants reference Lytics audiences:

- `VARIANTS_REFERRED_LYTICS_AUDIENCE_DOES_NOT_EXIST` – variants still contain Lytics audience UIDs
- `INSUFFICIENT_VARIANTS_TO_ACTIVATE` – all variants are removed, leaving 0 variants

**Root cause:** Lytics audiences are skipped during audience import (they can’t be created via API). Experience versions still send `lyticsAudiences` in the payload, but `lookUpAudiences` only processed `audiences`, so invalid Lytics references were never filtered.

## Solution

1. **audiences-helper.ts**
   - Process `lyticsAudiences` in `SegmentedVariant` the same way as `audiences`
   - Remove audience UIDs that are not in the mapping (same logic as `audiences`)
   - Remove variants when both `audiences` and `lyticsAudiences` are empty after filtering

2. **experiences.ts**
   - Only add versions to `versionMap` when they have at least one valid variant
   - Skip version import and log a warning when all variants become invalid

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| Variant with CMS audiences only | Unchanged | Unchanged |
| Variant with CMS + Lytics audiences | API error | Lytics refs removed; variant kept |
| Variant with only Lytics audiences | API error | Variant removed; version skipped if 0 variants |


sunil-lakshman (approved):
LGTM

harshithad0703 (approved):

aman19K (approved):

PR #2465, opened 2026-03-13
Enh/dx 5343

sunil-lakshman:


shafeeqd959 (commented):
version bump needed in plugins which are making use of utilities, except core contentstack package.

core contentstack package versions bump should be in a separate PR

cs-raj (approved):

harshitha07d (approved):

PR #2713, opened 2026-08-18
 Revert console transport override for error/warn levels

harshitha07d:
## Summary

Reverts the logger override in `cli-utilities` that forced `error`/`warn` levels to always reach the console (bypassing `showConsoleLogs`). Console transport is now added solely based on `showConsoleLogs`, restoring the original behavior.

## Change

`packages/contentstack-utilities/src/logger/logger.ts` — the console transport is pushed only when `showConsoleLogs` is true, dropping the `isErrorOrWarn` bypass.

## Context

The dropped override was the utilities-level approach to keeping failures visible in progress mode. That responsibility is instead handled at the plugin layer — see the companion bulk-operations fix (contentstack/cli-plugins#352), which surfaces failures via `catch()` without changing `cli-utilities`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


harshitha07d:
errors msgs were updated in cli-plugins

PR #2722, opened 2026-09-11
fix(deps): declare tslib as a runtime dependency

cs-raj:
### Problem

All five packages in this repo compile with `importHelpers: true` and
`target: es2017`, so `tsc` emits real `require("tslib")` calls into `lib/`.
None of them declared `tslib` in `dependencies`:

| Package | Files in `lib/` with `require("tslib")` | `tslib` declared in |
|---|---|---|
| `@contentstack/cli-utilities` | 28 | nowhere — not even devDependencies |
| `@contentstack/cli` | 4 | devDependencies only |
| `@contentstack/cli-auth` | 0 | nowhere |
| `@contentstack/cli-command` | 0 | nowhere |
| `@contentstack/cli-config` | 0 | nowhere |

`devDependencies` are not installed for consumers, so for packaging purposes the
first two rows are both undeclared.

`.npmrc` in this repo sets `shamefully-hoist=true`, so `tslib` is always present
in the hoisted root during development and always resolves.

On an install that does not hoist, it fails immediately. Reproduced against the
published `2.0.0` packages, in an empty directory with `hoist=false`:

```
pnpm add @contentstack/cli@2.0.0
csdx --help
```

```
Error: Unable to load configured help class "./lib/help.js", failed with message:
[MODULE_NOT_FOUND] require failed to load .../@contentstack/cli/lib/help.js:
Cannot find module 'tslib'
```

Exit code 1, before any help output. `lib/help.js` is this package's configured
`oclif.helpClass` and it needs `__importStar`, so the first thing the CLI loads
is a file that cannot resolve.

The failure is not limited to `--help`. Requiring every shipped module in that
same tree, **352 of 485 fail to load**, every one of them on `tslib`.
`cli-utilities` accounts for much of that spread: 24 of the 25 packages across
this repo and `cli-plugins` declare `@contentstack/cli-utilities` as a
dependency (the exception being `cli-utilities` itself), so its 28
missing-helper files also break packages that emit no helpers of their own.

### Fix

- Declare `tslib: ^2.8.1` in `dependencies` for all five packages. `^2.8.1` is
  the version used by every one of the six pre-existing `tslib` declarations
  across this repo and `cli-plugins`, so no new convention is introduced.
- `@contentstack/cli` already had `tslib` in `devDependencies` — moved rather
  than duplicated.
- `cli-auth`, `cli-command` and `cli-config` emit zero helpers today and are
  correct as shipped. They are included deliberately: all three set
  `importHelpers: true`, so the absence of helpers is a property of what their
  source currently happens to use, not a guarantee. Declaring `tslib` in all
  five makes the rule uniform and mechanically checkable — *a package that sets
  `importHelpers` declares `tslib`* — rather than leaving it to be rediscovered
  per package.

### Verification

All 25 packages across this repo and `cli-plugins` were packed through the real
`prepack` path and the tarballs installed into an empty directory with
`hoist=false`, with every `@contentstack/*` overridden to its local tarball so
nothing resolved to a published version:

| | Published `2.0.0` | This branch + the matching `cli-plugins` branch |
|---|---|---|
| `csdx --help` | exit 1, `Cannot find module 'tslib'` | exit 0 |
| Shipped modules that load | 133 / 485 | 486 / 486 |

Because the two repos' packages depend on each other, that result validates this
change together with the matching `cli-plugins` change; it is not a measurement
of this branch in isolation.

Also confirmed on this branch:

- All five packages resolve `tslib` from their own `node_modules`, not a hoisted
  root.
- `pnpm-lock.yaml` diff is 15 lines and touches nothing but `tslib`.
- No packed tarball manifest contains `workspace:*`.
- `pnpm run build` exits 0.


amit-kanswal-cs (approved):
LGTM

